PR #137, opened 2026-03-07
fix: Ctrl+N/P history navigation + approval window positioning (#134, #135)

### LINKED ISSUE #134: Enable Ctrl+N and Ctrl+P for command history navigation
---
name: Bug Report
about: Something isn't working right
labels: ["bug"]
---

## What happened?
Command history navigation does not support Ctrl+N and Ctrl+P, which is standard in many terminal interfaces.

## Steps to reproduce
Open the terminal, type some commands, then attempt to navigate history using Ctrl+N/Ctrl+P.

## Error output
N/A

## Environment
- **Koda version:** Unknown
- **OS:** Unknown
- **Provider/model:** Unknown
- **Mode:** Normal

## Relevant files
src/tui_app.rs, src/tui_commands.rs

## Additional context
Adding this support will improve user experience for those coming from other shell environments.

### LINKED ISSUE #135: Approval window rendering position
---
name: Bug Report
about: Something isn't working right
labels: ["bug"]
---

## What happened?
The inline approval window is rendering at the top of the terminal area instead of immediately above the prompt line.

## Steps to reproduce
Run any tool that requires approval.

## Error output
N/A (UI positioning issue)

## Environment
- **Koda version:** Unknown
- **OS:** Unknown
- **Provider/model:** Unknown
- **Mode:** Normal

## Relevant files
src/tui_app.rs, src/widgets/approval.rs

## Additional context
This makes it difficult to associate the approval request with the action being performed when there is existing output.

lijunzh:
### Ctrl+N/Ctrl+P for command history (#134)
Added `Ctrl+P` (previous) and `Ctrl+N` (next) as aliases for Up/Down arrow keys when navigating command history. This matches standard readline/emacs-style keybindings.

### Approval window rendering position (#135)
Fixed the approval options menu rendering inside the ratatui viewport area (overlapping the status/input bar) instead of immediately above it.

**Root cause:** After `insert_before()` calls emit the approval header, the cursor sits at the viewport row. The crossterm-based options menu then wrote over the viewport.

**Fix:** Scroll the terminal to make room for the options menu before rendering, then move the cursor back up — matching the pattern used by `select_inline()`.

## Testing
- `cargo fmt --all --check` ✅
- `cargo clippy --workspace -- -D warnings` ✅
- `cargo test --workspace --features koda-core/test-support` ✅

Closes #134
Closes #135

### DIFF
--- a/koda-cli/src/tui_app.rs
+++ b/koda-cli/src/tui_app.rs
@@ -622,48 +622,46 @@ pub async fn run(
                                                 if preview.is_some() {
                                                     renderer.preview_shown = true;
                                                 }
-                                                // Inline approval — stays in raw mode
+                                                // Inline approval — uses crossterm direct writes
                                                 let decision = crate::widgets::approval::prompt_approval(
-                                                    &mut terminal,
                                                     &tool_name,
                                                     &detail,
                                                     preview.as_ref(),
                                                     whitelist_hint.as_deref(),
                                                 );
+                                                // Resync ratatui viewport after crossterm writes
+                                                crossterm_events = EventStream::new();
+                                                terminal = init_terminal(viewport_height)?;
                                                 let _ = cmd_tx
                                                     .send(EngineCommand::ApprovalResponse { id, decision })
                                                     .await;
                                             }
                                             UiEvent::Engine(EngineEvent::LoopCapReached { cap, recent_tools }) => {
-                                                // Show cap info above viewport
-                                                tui_output::emit_blank(&mut terminal);
-                                                tui_output::emit_line(
-                                                    &mut terminal,
-                                                    Line::from(vec![
-                                                        Span::raw("  "),
-                                                        Span::styled(
-                                                            format!("\u{26a0} Hard cap reached ({cap} iterations)"),
-                                                            Style::default().fg(Color::Yellow),
-                                                        ),
-                                                    ]),
-                                                );
+                                                // Show cap info via crossterm (matches approval widget path)
+                                                tui_output::write_blank();
+                                                tui_output::write_line(&Line::from(vec![
+                                                    Span::raw("  "),
+                                                    Span::styled(
+                                                        format!("\u{26a0} Hard cap reached ({cap} iterations)"),
+                                                        Style::default().fg(Color::Yellow),
+                                                    ),
+                                                ]));
                                                 for name in &recent_tools {
-                                                    tui_output::emit_line(
-                                                        &mut terminal,
-                                                        Line::from(vec![
-                                                            Span::raw("    "),
-                                                            Span::styled(format!("\u{25cf} {name}"), Style::default().fg(Color::DarkGray)),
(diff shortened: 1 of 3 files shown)


lijunzh:
## Root Cause Analysis: Approval menu renders at top of terminal

### Problem
The approval options menu renders at the **top** of the visible terminal instead of immediately above the ratatui viewport (input/status bar at the bottom).

### Root Cause: Cursor position mismatch after `insert_before()`

The approval widget in `koda-cli/src/widgets/approval.rs` mixes two rendering systems in the same operation:

1. **ratatui `insert_before()`** (via `emit_line()`) — for the tool detail text and diff preview
2. **crossterm raw cursor manipulation** — for the interactive options menu

After `insert_before()` returns, the physical cursor position is **not** at the viewport. Tracing through ratatui 0.29's `insert_before_scrolling_regions()` implementation:

- It uses `scroll_region_up()` to shift content above the viewport
- Then `draw_lines_over_cleared()` renders the inserted content using `backend.draw()`, which uses absolute `MoveTo(x, y)` positioning
- **The function returns without repositioning the cursor to the viewport.** The cursor is left wherever the last `draw()` call placed it — typically just above the viewport area, but at an unpredictable column/row.

When the approval widget then executes its scroll code:
```rust
for _ in 0..total_lines {
    crossterm::execute!(stdout, Print("\n")).ok();
}
crossterm::execute!(stdout, cursor::MoveUp(total_lines)).ok();
```

The newlines and `MoveUp` are relative to wherever `insert_before` left the cursor, not relative to the viewport. This places the options menu at an unintended position (near the top of the terminal).

### Why `select_inline()` works but `prompt_approval()` doesn't

`select_inline()` (in `select_menu.rs`) is only called from slash commands, which use the **crossterm-only** rendering path (`write_line()`). The cursor position is always correct because crossterm manages it consistently. The approval widget is the only place that mixes `insert_before()` with crossterm cursor manipulation.

### Additional bug (secondary)

`render_options()` uses `MoveUp(height)` where `height = options.len() + 2`, but the cursor only moves down `height - 1` times during rendering (title MoveDown + N option MoveDowns; hint line has no MoveDown). This causes a 1-line upward drift on every re-render. Compare with `select_menu.rs:render_inline()` which correctly uses `MoveUp(height - 1)`.

### Proposed Fix

Stop mixing rendering systems in the approval widget. Two options:

**Option A (recommended):** Use `write_line()` (crossterm path) for the detail text instead of `emit_line()` (ratatui `insert_before`). The approval widget temporarily takes over rendering anyway. After approval completes, reinitialize the terminal at the call site (`init_terminal()`) to resync ratatui's viewport tracking — matching the pattern used after slash commands.

**Option B:** After the `emit_line()` calls, query the actual cursor position with `crossterm::cursor::position()` and use absolute `MoveTo` positioning instead of relative `MoveUp`. More fragile since `cursor::position()` requires a DSR query that can conflict with the event stream.

Issue #49, opened 2026-03-05
fix: case-insensitive tool name matching for small models


lijunzh:
## Problem

Small models (e.g., `qwen2.5-coder-7b-instruct`) sometimes send tool names in lowercase (`bash`, `read`, `list`) instead of the expected PascalCase (`Bash`, `Read`, `List`). The tool dispatcher does exact string matching, so these calls fail with `Unknown tool: bash`.

The approval system has the same issue — `READ_ONLY_TOOLS` uses PascalCase, so a lowercase tool name bypasses auto-approval and triggers a confirmation prompt for read-only tools.

## Examples

```
● Tool bash  Execute: bash
│ Error: Unknown tool: bash

● Tool read  Execute: read
🐻 Confirm action?   ← Read should be auto-approved!
```

## Fix

1. **Tool dispatcher** (`tools/mod.rs`): normalize tool name to PascalCase before matching
2. **Approval** (`approval.rs`): case-insensitive comparison for `READ_ONLY_TOOLS`
3. **Inference** (`inference.rs`): normalize `tc.function_name` early so all downstream code sees PascalCase

Simplest approach: add a `normalize_tool_name()` function that maps common lowercase variants to PascalCase, called once when tool calls are received from the LLM.

PR #549, opened 2026-03-24
fix: normalize tool names from model output to canonical PascalCase (#548)

### LINKED ISSUE #548: bug: OpenAI-compatible providers return raw tool names, causing  with LM Studio
## Summary

Using an OpenAI-compatible provider (repro from LM Studio with `zai-org/glm-4.6v-flash`), the model can emit lowercase or snake_case tool names such as `list` / `list_files` instead of the canonical PascalCase names (`List`, `Read`, etc.). Koda currently forwards the raw tool name into dispatch, which fails with:

```text
Error: Unknown tool: list
```

This can also trigger loop detection because the model keeps retrying the same invalid tool call.

## Repro

Provider: `lm-studio`
Model: `zai-org/glm-4.6v-flash`

Prompt / command:

```text
ls
```

Observed transcript excerpt:

```text
● list
│ Error: Unknown tool: list
...
⚠ Loop detected: list is repeating with identical arguments.
```

The user even corrected the model (`the tool name is List instead of list`) and it still continued emitting `list`.

## Expected

Tool calls from providers should be normalized to canonical internal names before approval/dispatch, so common variants like these work:

- `list` -> `List`
- `read` -> `Read`
- `bash` -> `Bash`
- `list_files` -> `List`
- `read_file` -> `Read`
- etc.

At minimum, the dispatcher should not fail on common lowercase / snake_case variants coming from OpenAI-compatible providers.

## Actual

The tool registry / dispatcher only matches canonical PascalCase names:

- `koda-core/src/tools/mod.rs`
  - `ToolRegistry::execute()` matches on `"List"`, `"Read"`, etc.
- `koda-core/src/approval.rs`
  - `check_tool()` / `resolve_effect()` also rely on canonical names

Meanwhile, provider code returns tool names as-is:

- `koda-core/src/providers/openai_compat.rs`
  - streaming parser accumulates raw names into `ToolCall.function_name`
  - non-streaming `chat()` also uses `tc.function.name` directly
- `koda-core/src/inference.rs`
  - `collect_stream()` forwards `StreamChunk::ToolCalls(tcs)` unchanged

There also appears to be naming drift in tests/docs:

- `koda-core/tests/new_tools_test.rs` asserts built-ins are PascalCase
- `koda-core/src/providers/stream_collector.rs` has a Gemini parser test fixture expecting `list_files` / `read_file`

## Related

This seems related to closed issue #49 (`fix: case-insensitive tool name matching for small models`), but the current code path still appears to pass raw names through unchanged for OpenAI-compatible / LM Studio flows.

## Suggested fix

Add a single canonical normalization layer and use it early when tool calls are received from the model, e.g.:

1. Introduce `normalize_tool_name()` in a shared location
2. Call it when constructing `ToolCall` values in provider parsers and/or in `collect_stream()` before any downstream logic
3. Consider supporting both lowercase and known snake_case aliases
4. Add regression tests for:
   - OpenAI-compatible streaming tool calls with `list`
   - OpenAI-compatible streaming tool calls with `list_files`
   - approval classification on normalized names
   - dispatcher execution after normalization

## Why this matters

This is provider/model compatibility breakage, not just prompt quality. Some smaller or local models will confidently emit lowercase/snake_case names even when the tool schema uses PascalCase. Without normalization, local-provider usability degrades badly and can fall into useless retry loops.

lijunzh:
## Summary

Fixes #548 — Models (especially local/small ones via LM Studio) emit lowercase or snake_case tool names (`list`, `list_files`, `read_file`) instead of PascalCase (`List`, `Read`). This caused `Unknown tool` errors and infinite retry loops.

## Root cause

The dispatcher (`tools/mod.rs`) and approval system (`approval.rs`) do exact PascalCase matching, but the OpenAI-compatible provider path returns raw model-emitted names with no normalization. Issue #49 identified this class of bug but the fix was incomplete.

## What this PR does

**New module: `koda-core/src/tool_normalize.rs`** (315 lines)

- `LazyLock<HashMap>` alias map covering:
  - Automatic lowercase mappings for all 20 canonical tools
  - Common snake_case variants (`list_files` → `List`, `read_file` → `Read`, etc.)
  - Short aliases models hallucinate (`ls` → `List`, `rm` → `Delete`, `rg` → `Grep`)
- `normalize_tool_name()` — single name normalization with fast-path for canonical names
- `normalize_tool_calls()` — batch normalization for tool call vectors

**Modified: `koda-core/src/inference.rs`** (4-line change)

- Normalize tool calls once after `collect_stream()` returns, before dispatch/approval/persistence
- Single boundary normalization point (DRY, not scattered across consumers)

## Design decisions

| Decision | Rationale |
|----------|-----------|
| **Keep PascalCase** | Already embedded in 20 tools, dispatch, approval, tests. Would be massive refactor. |
| **Normalize at boundary** | Single point in inference.rs, not at every consumer. DRY + SRP. |
| **Unknown passthrough** | Unknown names pass through so dispatcher gives clear errors. |
| **Static alias map** | `LazyLock<HashMap>` — O(1) lookup, built once, zero runtime cost after init. |

## Tests

**8 unit tests** (in `tool_normalize.rs`):
- Canonical passthrough, lowercase, snake_case, short aliases, mixed case, unknowns, batch, sync check

**3 integration tests** (in `tests/tool_normalize_test.rs`):
- Lowercase names routable through dispatcher
- Snake_case names routable through dispatcher
- Normalizer CANONICAL list synced with registry

All existing tests pass (tool_wiring_test, new_tools_test).

## Repro from the issue

```
Provider: lm-studio (zai-org/glm-4.6v-flash)
> ls
● list
│ Error: Unknown tool: list
⚠ Loop detected: list is repeating with identical arguments.
```

After this fix: `list` → `List` → dispatched correctly ✅

### DIFF
--- a/koda-core/src/inference.rs
+++ b/koda-core/src/inference.rs
@@ -627,8 +627,11 @@ pub async fn inference_loop(ctx: InferenceContext<'_>) -> Result<()> {
 
         let full_text = stream_result.text;
         // Normalize tool names from model output to canonical PascalCase (#548).
-        // Models (especially local/small ones) may emit lowercase or snake_case
-        // names; this must happen before dispatch, approval, or persistence.
+        // Models (especially local/small ones via OpenAI-compat APIs) may emit
+        // lowercase or snake_case names ("list", "read_file"). This runs for all
+        // providers — the canonical fast-path is a single HashMap lookup — and
+        // must happen here (not in providers) so dispatch, approval, loop guard,
+        // undo, and persistence all see consistent canonical names.
         let tool_calls = crate::tool_normalize::normalize_tool_calls(stream_result.tool_calls);
         let usage = stream_result.usage;
         let char_count = stream_result.char_count;
--- a/koda-core/src/inference.rs
+++ b/koda-core/src/inference.rs
@@ -626,7 +626,10 @@ pub async fn inference_loop(ctx: InferenceContext<'_>) -> Result<()> {
         }
 
         let full_text = stream_result.text;
-        let tool_calls = stream_result.tool_calls;
+        // Normalize tool names from model output to canonical PascalCase (#548).
+        // Models (especially local/small ones) may emit lowercase or snake_case
+        // names; this must happen before dispatch, approval, or persistence.
+        let tool_calls = crate::tool_normalize::normalize_tool_calls(stream_result.tool_calls);
         let usage = stream_result.usage;
         let char_count = stream_result.char_count;
 
--- a/koda-core/src/lib.rs
+++ b/koda-core/src/lib.rs
@@ -66,6 +66,8 @@ pub mod skills;
 pub mod sub_agent_cache;
 /// Tool dispatch — routes tool calls from inference to the registry.
 pub mod tool_dispatch;
+/// Tool name normalization — maps model-emitted variants to canonical PascalCase.
+pub mod tool_normalize;
 /// Tool registry, definitions, execution, and path safety.
 pub mod tools;
 /// Token-safe output truncation.
--- a/koda-core/src/tool_normalize.rs
+++ b/koda-core/src/tool_normalize.rs
@@ -244,7 +244,15 @@ mod tests {
     fn ambiguous_names_not_mapped() {
         // These could plausibly map to multiple tools.
         // Better to surface "Unknown tool" than silently misroute.
-        for name in ["search", "execute", "exec", "patch", "terminal", "find_files", "fetch"] {
+        for name in [
+            "search",
+            "execute",
+            "exec",
+            "patch",
+            "terminal",
+            "find_files",
+            "fetch",
+        ] {
             let result = normalize_tool_name(name);
             assert_eq!(
                 result, name,
@@ -323,8 +331,7 @@ mod tests {
 
     #[test]
     fn all_alias_targets_are_canonical() {
-        let canonical_set: std::collections::HashSet<&str> =
-            CANONICAL.iter().copied().collect();
+        let canonical_set: std::collections::HashSet<&str> = CANONICAL.iter().copied().collect();
         for (alias, &target) in ALIASES.iter() {
             assert!(
                 canonical_set.contains(target),
--- a/koda-core/src/tool_normalize.rs
+++ b/koda-core/src/tool_normalize.rs
@@ -1,0 +1,315 @@
+//! Tool name normalization — maps model-emitted variants to canonical PascalCase.
+//!
+//! Models (especially smaller/local ones via OpenAI-compatible APIs) sometimes
+//! emit tool names in lowercase (`list`, `read`) or snake_case (`list_files`,
+//! `read_file`) instead of the canonical PascalCase (`List`, `Read`). This
+//! module provides a single normalization point at the API boundary so all
+//! downstream code (dispatch, approval, loop guard, undo) sees canonical names.
+//!
+//! ## Design
+//!
+//! - Normalization is applied *once*, in `inference.rs`, after collecting
+//!   the streamed response — before dispatch, approval, or persistence.
+//! - Unknown names pass through unchanged so the dispatcher can surface
+//!   a clear `Unknown tool` error.
+//! - The alias map covers lowercase, snake_case, and camelCase variants.
+//!
+//! See: <https://github.com/lijunzh/koda/issues/548>
+//!      <https://github.com/lijunzh/koda/issues/49>
+
+use crate::providers::ToolCall;
+use std::collections::HashMap;
+use std::sync::LazyLock;
+
+/// All canonical (PascalCase) built-in tool names.
+const CANONICAL: &[&str] = &[
+    "ActivateSkill",
+    "AstAnalysis",
+    "Bash",
+    "Delete",
+    "Edit",
+    "EmailRead",
+    "EmailSearch",
+    "EmailSend",
+    "Glob",
+    "Grep",
+    "InvokeAgent",
+    "List",
+    "ListAgents",
+    "ListSkills",
+    "MemoryRead",
+    "MemoryWrite",
+    "Read",
+    "RecallContext",
+    "WebFetch",
+    "Write",
+];
+
+/// Static alias map: lowercased variant → canonical name.
+///
+/// Built once on first access.  Includes:
+/// - lowercase of every canonical name  (`"list"` → `"List"`)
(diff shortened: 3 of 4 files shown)


lijunzh:
## Code Review — Normalization layer vs. snake_case rename

Good fix for a real usability pain point. Here's my analysis of the core architectural question and specific concerns with the PR.

---

### Should we rename to snake_case instead?

**Short answer: No. The normalization layer is the right call.**

Moving to snake_case (`list`, `read`, `memory_read`, etc.) would eliminate the root mismatch for well-behaved models, but it doesn't actually solve the problem:

1. **Small models will still hallucinate variants.** Even with snake_case canonical names, a model might still emit `list_all_files`, `ls`, `listFiles`, or `run_command`. The normalization layer is inherently more robust — it handles arbitrary deviations regardless of what canonical form we pick.

2. **The rename blast radius is large.** 20 tools × dispatch match arms + approval match arms + test assertions + documentation. `new_tools_test.rs` has an explicit `test_no_underscores_in_tool_names` guard — that test exists for a reason. Reversing this convention mid-project adds churn without eliminating the need for normalization.

3. **PascalCase has a distinct advantage.** Names like `List`, `Read`, `Bash` are unlikely to conflict with common Unix commands or environment variables at the schema level. `list` and `read` are used in many contexts; using them as canonical names creates ambiguity about intent in logs and approval prompts.

4. **The fix is already additive and isolated.** `tool_normalize.rs` is 315 lines in a new module with zero changes to existing dispatch/approval code. The 4-line change in `inference.rs` is the smallest possible touch point.

---

### Concerns with the PR as written

**1. Ambiguous "hallucination" aliases are risky**

The aliases that worry me are ones where the name is genuinely ambiguous:

```rust
m.insert("search".into(), "Grep");   // could mean Glob
m.insert("execute".into(), "Bash");  // too generic
m.insert("exec".into(), "Bash");     // ditto
m.insert("patch".into(), "Edit");    // patch ≠ edit semantically
m.insert("terminal".into(), "Bash"); // stretch
m.insert("find_files".into(), "Glob"); // but "find" → ? (not mapped)
```

The problem: when `search` silently routes to `Grep` but the model intended a glob pattern search, you get wrong behavior with no error. That's worse than the current `Unknown tool` failure, which at least surfaces the problem. I'd suggest restricting to unambiguous aliases only — things where there's only one plausible canonical target.

**2. `CANONICAL.contains(&name)` is a linear scan**

```rust
if CANONICAL.contains(&name) {
    return name.to_string();
}
```

`CANONICAL` is a `&[&str]` so this is O(n). With 20 elements it's fine, but since you're already building a `HashMap`, you could just include the canonical names as self-mapping entries in `ALIASES` and do one O(1) lookup instead. This also simplifies the code path.

**3. `InvokeAgent` missing from integration test**

`normalized_lowercase_names_are_routable` tests `"invokeagent"` is not in its list, despite being in CANONICAL. Minor — just an oversight.

**4. Normalization placement is correct but worth documenting**

Normalizing in `inference.rs` after `collect_stream()` means it runs for all providers, including Anthropic which already returns correct names. This is fine (canonical fast-path is cheap), but a brief comment at the call site saying *why* it's here (not in provider code) would help future maintainers understand that this is intentionally a single unified boundary, not a per-provider concern.

**5. Alias map maintenance is an ongoing responsibility**

The `normalizer_covers_all_registry_tools` integration test catches the critical drift (new tool added to registry but not CANONICAL), which is the right guardrail. But the inverse — a new tool added to CANONICAL without updating aliases — won't be caught. This is low priority since the auto-generated lowercase mappings cover the minimum viable case.

---

### Summary

| Question | Recommendation |
|----------|---------------|
| Normalization layer or rename? | **Normalization layer** — rename doesn't eliminate the need for it and has large blast radius |
| Is this PR the right approach? | **Yes**, with minor cleanup |
| Ambiguous aliases (`search`, `exec`, etc.)? | **Remove or comment with warning** — silent misrouting is worse than an error |
| Integration test gap? | Add `invokeagent` to `normalized_lowercase_names_are_routable` |
| Fast-path linear scan? | Low priority, but could unify into a single map lookup |

The normalization layer at the `inference.rs` boundary is the correct design. It's DRY, it has a clear single responsibility, and it doesn't require touching any of the 20 tool implementations. The main thing I'd tighten before merging is the ambiguous alias set — keeping only mappings where there's one unambiguous canonical target.

lijunzh:
## Follow-up review — ready to merge ✓

Checked the two commits pushed after the initial review (`1d08718`, `6617016`). All five concerns were addressed:

---

**1. Ambiguous aliases — fully resolved**

Removed exactly the aliases flagged as risky: `search`, `execute`, `exec`, `patch`, `terminal`, `find_files`, `fetch`. The doc comment on `ALIASES` now states the policy explicitly:

> *Only unambiguous aliases are included. If a name could plausibly map to more than one tool, it is intentionally omitted — surfacing an `Unknown tool` error is better than silently misrouting.*

Bonus: a new `ambiguous_names_not_mapped` test asserts all seven removed names pass through unchanged. This locks the policy in code so it can't regress silently.

**2. O(n) linear scan — eliminated**

The `CANONICAL.contains(&name)` pre-check is gone. Canonical names are now self-mapped in the auto-generation loop (`"list" → "List"`, etc.), so `normalize_tool_name` is a single `ALIASES.get(&lower)` O(1) lookup on every path. The function body went from 8 lines to 4.

**3. Missing `invokeagent` in integration test — fixed**

`"invokeagent"` added to `normalized_lowercase_names_are_routable`. All 17 lowercase variants of built-in tools are now covered.

**4. Inference.rs comment — improved**

Expanded from 2 lines to 5, now explicitly explains why normalization lives here (not in provider code) and enumerates the downstream systems that depend on seeing canonical names: dispatch, approval, loop guard, undo, persistence.

**5. Alias target integrity — new test added**

`all_alias_targets_are_canonical` iterates the entire `ALIASES` map and asserts every target value exists in `CANONICAL`. Catches the inverse drift (bad target added to aliases) that `normalizer_covers_all_registry_tools` doesn't cover.

---

**Bonus cleanup noticed:** The explicit entries for `"invokeagent"`, `"activateskill"`, `"astanalysis"`, `"recallcontext"`, `"webfetch"` were removed from the alias section — they were redundant with the auto-generated lowercase loop. The explicit section now contains only genuinely non-obvious aliases (snake_case compounds, Unix shorthands). Cleaner signal-to-noise.

**No remaining concerns.** The implementation is minimal, well-tested, and correctly bounded. Approve to merge.

PR #582, opened 2026-04-01
security: extract should_skip_tls_verification() for clarity

### LINKED ISSUE #575: 🔴 [CodeQL] Disabled TLS certificate check in providers/mod.rs
## Code Scanning Alert #13 — Severity: **HIGH**

**File:** `koda-core/src/providers/mod.rs` (line 199)
**Rule:** `rust/disabled-certificate-check`

### Description
CodeQL flagged `danger_accept_invalid_certs(true)` as disabling TLS certificate validation, which can expose the application to man-in-the-middle attacks.

### Current Behavior
The code already gates this behind `KODA_ACCEPT_INVALID_CERTS` env var AND checks that the URL is localhost — so the risk is mitigated in practice. However, CodeQL still flags it because the API call itself is unconditionally dangerous.

### Suggested Fix Options
1. **Suppress with justification** — Add a CodeQL suppression comment explaining the localhost-only guard. This is a reasonable choice since the existing guard is solid.
2. **Use a dedicated local-only client** — Build a separate `reqwest::Client` for local providers so the dangerous call is more clearly isolated and documented.

### Action Items
- [ ] Evaluate whether suppression or refactoring is preferred
- [ ] Add inline documentation explaining the security rationale
- [ ] Verify alert #13 is resolved after the fix

### References
- [Code Scanning Alert #13](https://github.com/lijunzh/koda/security/code-scanning/13)

lijunzh:
Refactor the TLS bypass logic in `providers/mod.rs` into a dedicated `should_skip_tls_verification()` function with full safety documentation.

**What changed:**
- Extracted the env-var + localhost check into a named function with doc comments
- Added `lgtm` suppression on the `danger_accept_invalid_certs` call  
- No behavioral change — same two-condition guard (env var + localhost only)

**Security rationale:**
The TLS bypass exists for local LLM servers (LM Studio, Ollama, vLLM) that use self-signed certs. It requires both:
1. `KODA_ACCEPT_INVALID_CERTS=1` env var
2. Provider URL pointing to localhost/127.0.0.1/::1

Remote providers NEVER skip TLS verification.

Closes #575

### DIFF
--- a/koda-core/src/providers/mod.rs
+++ b/koda-core/src/providers/mod.rs
@@ -147,6 +147,33 @@ fn is_localhost_url(url: &str) -> bool {
     lower.contains("://localhost") || lower.contains("://127.0.0.1") || lower.contains("://[::1]")
 }
 
+/// Check whether TLS certificate verification should be skipped.
+///
+/// # Safety rationale
+///
+/// This returns `true` ONLY when both conditions are met:
+/// 1. `KODA_ACCEPT_INVALID_CERTS` env var is explicitly set to `"1"` or `"true"`
+/// 2. The provider URL points to localhost (127.0.0.1, ::1, or `localhost`)
+///
+/// This is needed for local LLM servers (LM Studio, Ollama, vLLM) that use
+/// self-signed certificates. Remote providers NEVER skip TLS verification.
+fn should_skip_tls_verification(base_url: Option<&str>) -> bool {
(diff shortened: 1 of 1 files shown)


lijunzh:
Closing: the original code is already safe (double-guarded by env var + localhost check). Dismissed CodeQL alert #13 as false positive instead. The refactor moved the line and caused CodeQL to flag it as a new alert — net negative.

PR #659, opened 2026-04-04
fix: include full_content in load_messages_before query

### LINKED ISSUE #658: bug: load_messages_before query missing full_content column
## Problem

`load_messages_before` in `db/mod.rs` (line ~218) selects:
```sql
SELECT id, session_id, role, content, tool_calls, tool_call_id,
       prompt_tokens, completion_tokens,
       cache_read_tokens, cache_creation_tokens, thinking_tokens,
       created_at
FROM messages ...
```

But `load_context` and `load_all_messages` both include `full_content`:
```sql
SELECT id, session_id, role, content, full_content, tool_calls, ...
```

`full_content` is used for Bash tool results that have a summary (shown to model) + full output (stored separately for RecallContext search).

## Impact

`load_messages_before` is used for virtual scroll pagination. The missing column means `full_content` is always `None` for paginated messages, which could cause issues if any code path relies on it during scroll-back.

Since `MessageRow` derives `FromRow` and `full_content` is `Option<String>`, sqlx silently sets it to `None` rather than erroring — so this has been silently wrong.

## Fix

Add `full_content` to the SELECT in `load_messages_before`.

lijunzh:
The `load_messages_before` function was missing the `full_content` column in its SELECT statement, while both `load_context` and `load_all_messages` include it. This caused `full_content` to be silently set to None for paginated messages during virtual scroll pagination.

Fixes #658

### DIFF
--- a/koda-core/src/db/mod.rs
+++ b/koda-core/src/db/mod.rs
@@ -216,7 +216,7 @@ impl Database {
         limit: i64,
     ) -> Result<Vec<Message>> {
         let rows: Vec<MessageRow> = sqlx::query_as(
-            "SELECT id, session_id, role, content, tool_calls, tool_call_id,
+            "SELECT id, session_id, role, content, full_content, tool_calls, tool_call_id,
                     prompt_tokens, completion_tokens,
                     cache_read_tokens, cache_creation_tokens, thinking_tokens
              FROM messages


lijunzh:
This is a PR created by koda with local model (qwen3.5-122b-a10b). Claude opus 4.6 reviewed before merge.

Issue #637, opened 2026-04-04
v0.2.x: CC features Koda doesn't have yet (future)


lijunzh:
# v0.2.x: CC features Koda doesn't have yet

> Features Claude Code has that Koda does not. These are **new capabilities**, not quality gaps in existing features. Tracked for v0.2.x releases after the v0.2.0 quality milestone.

## Context

Full architectural comparison against Claude Code v2.1.88. Koda's architecture already matches CC on core design (inference loop, tool parallelization, compaction, sub-agents). These are additive features that would expand Koda's capability surface.

**Note**: Koda is ahead of CC in some areas: multi-provider support (Anthropic + OpenAI + Gemini), SQLite storage, git worktree isolation for sub-agents, and file-level undo.

---

### v0.2.1 candidates (high value, bounded scope)

- [ ] **MCP support** — Model Context Protocol client integration. CC has 23 files (427KB) for MCP. Enables third-party tool servers (databases, APIs, browser automation). This is the highest-leverage missing feature — it's an extensibility multiplier.
  - CC source: `src/services/mcp/`

- [ ] **Plan mode** — `EnterPlanMode` / `ExitPlanMode` tools. Model transitions to read-only exploration mode, designs an approach, gets user approval, then exits to implement. CC uses this as a permission mode switch (plan mode = read-only tools only).
  - Koda has a `plan` sub-agent but no mode switch on the main agent.
  - CC source: `src/tools/EnterPlanModeTool/`, `src/tools/ExitPlanModeTool/`

- [ ] **Tool result persistence to disk** — Large tool results saved to files, replaced in context with `<persisted-output>` stubs. Model can Read them back if needed. Preserves information while keeping context slim.
  - CC source: `src/utils/toolResultStorage.ts` (1,040 lines)

### v0.2.2 candidates (medium value)

- [ ] **Hooks system** — User-defined shell commands executed at lifecycle points (pre-tool, post-tool, pre-compact, post-compact, session-start). Enables CI integration, custom validation, audit logging.
  - CC source: `src/utils/hooks.ts` (5,022 lines)
  - Start small: pre-tool and post-tool hooks only

- [ ] **ToolSearch tool** — Model can discover available tools by searching descriptions. Useful when MCP adds many tools and the model needs to find the right one.
  - CC source: `src/tools/ToolSearchTool/` + `src/utils/toolSearch.ts` (1,349 lines total)

[...]

PR #665, opened 2026-04-04
feat: dual-threshold auto-compact (advisory at 70%, critical at 90%)

lijunzh:
## Summary

Replace single hard-coded compact threshold with two configurable thresholds to reduce the "surprise cliff" when context suddenly fills up.

## Problem

Previously, koda had a single auto-compact threshold at 90% (hard-coded in `inference_helpers.rs`). When hit, it aggressively summarized the oldest half of messages — a jarring loss of context in a single step. Additionally, `auto_compact_threshold` existed in `AgentConfig` (default 85%) but was **dead code** — never read by the inference loop.

## Solution: Dual-Threshold Compaction

| Threshold | Default | What happens |
|---|---|---|
| Advisory | 70% | Auto-compact (oldest half) |
| Warning | 80% | User notification only |
| Critical | 90% | Auto-compact (oldest half) |

Both thresholds use the **same compaction strategy** (summarize oldest half). The value of dual thresholds is *when* to compact, not *how* — by compacting early at 70%, we avoid the surprise cliff at 90%. If the advisory compact freed enough space, the session continues without ever hitting the critical threshold.

## Changes

- `compact.rs`: Remove mode complexity, single `COMPACT_FRACTION = 0.5`
- `config.rs`: Wire up `auto_compact_threshold` (was dead code!), add `advisory_compact_threshold`
- `inference.rs`: Check both thresholds, log which level triggered ("advisory" vs "critical")
- `inference_helpers.rs`: Replace `PREFLIGHT_COMPACT_THRESHOLD` with `DEFAULT_CRITICAL_COMPACT_THRESHOLD` and `DEFAULT_ADVISORY_COMPACT_THRESHOLD`

## Config

Both thresholds are configurable in agent JSON:

```json
{
  "auto_compact_threshold": 90,
  "advisory_compact_threshold": 70
}
```

Set either to `0` to disable that level.

## Tests

All existing tests pass. `cargo fmt` + `cargo clippy` clean.

Refs #637

### DIFF
--- a/koda-cli/src/tui_handlers_inference.rs
+++ b/koda-cli/src/tui_handlers_inference.rs
@@ -236,7 +236,6 @@ impl TuiContext {
             self.config.max_context_tokens,
             &self.config.model_settings,
             &self.provider,
-            koda_core::compact::CompactMode::Critical,
         )
         .await
         {
--- a/koda-cli/src/tui_handlers_inference.rs
+++ b/koda-cli/src/tui_handlers_inference.rs
@@ -236,6 +236,7 @@ impl TuiContext {
             self.config.max_context_tokens,
             &self.config.model_settings,
             &self.provider,
+            koda_core::compact::CompactMode::Critical,
         )
         .await
         {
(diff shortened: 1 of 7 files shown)


lijunzh:
Closing — dual-threshold sounds good in theory but isn't tested. CC's single-threshold behavior works well in practice. Deferring to future.

The dead-code fix (`auto_compact_threshold` not wired up) is minor and can be addressed separately if needed.

PR #780, opened 2026-04-09
docs(CLAUDE.md): document required gates rationale (Test + Docs only)

lijunzh:
Lint and Check are transitively enforced by `test`'s `needs: [lint, check]`. Requiring them explicitly as branch gates is redundant — the DAG does it.


lijunzh:
Superseded by #781 which folds this change in alongside the crates.io publish ordering fix.

Issue #823, opened 2026-04-11
bug(gemma 4): model stuck in thinking mode with continuous token generation but no output


lijunzh:
## Description

When using the new Google Gemma 4 model via LM Studio, Koda gets stuck in thinking mode without outputting any additional tokens. However, in LM Studio I can see that tokens are being generated continuously.

## Expected Behavior

Koda should display the streaming token output as it receives it from the model.

## Actual Behavior

- Koda appears frozen in "thinking" state
- No new tokens are displayed in the TUI
- LM Studio shows continuous token generation (confirming the model is working)

## Environment

- Model: Google Gemma 4 (via LM Studio)
- Provider: OpenAI-compatible (LM Studio)

## Possible Causes

1. Stream parsing issue: The SSE stream collector or chunk parser may not be handling Gemma 4's response format correctly
2. Tag filtering bug: stream_tag_filter.rs might be stripping all content thinking it's XML tool tags
3. Buffering problem: Tokens are being received but not flushed to the TUI
4. Model-specific format: Gemma 4 may use a different streaming format that isn't handled by the current OpenAI-compatible provider

## Steps to Reproduce

1. Configure LM Studio with Google Gemma 4 model
2. Select the model in Koda
3. Send any prompt
4. Observe: tokens generate in LM Studio but Koda stays frozen

## Additional Context

This appears to be a model-specific issue with Gemma 4's streaming format when served through LM Studio.

PR #832, opened 2026-04-11
fix: revert local model workarounds + adopt Gemini CLI loop detection (#831)

### LINKED ISSUE #830: question: is 5 consecutive tool-only responses the right threshold for tool suppression?
## Context

When a model produces 5 consecutive tool-only responses (no text to the user), koda suppresses tools for the next turn to force a text reply:

> *"Model produced 5 consecutive tool-only responses — suppressing tools for this turn to prompt a text reply."*

## Question

Is 5 the right threshold? A few considerations:

**Arguments for a larger threshold (e.g. 10–20):**
- Legitimate agentic tasks — multi-step file edits, code refactors, recursive searches — can naturally produce many back-to-back tool calls without any user-facing text. Suppressing tools at 5 may interrupt valid workflows mid-task.
- The suppression message seen in practice came from a model that was already in an infinite loop (370+ calls, see #826). At that scale, 5 vs. 10 makes no difference to the loop, but 5 *does* interrupt valid deep-tool tasks earlier than necessary.

**Arguments for keeping it at 5 (or going lower):**
- A low threshold is a cheap early warning for runaway local models (Gemma 4 via lm-studio, etc.) that don't respond to stop sequences.
- Suppressing tools once is low-cost and recoverable — the model can request tools again on the next turn if it needs them.

**Alternative approaches worth considering:**
- Make the threshold configurable per-agent or per-provider (local models may warrant a stricter cap than cloud providers).
- Distinguish between *diverse* consecutive tool calls (plausibly intentional) and *repeated identical* tool calls (likely a loop) — the repeat-detection in `loop_guard.rs` already does something like this; the two mechanisms could be unified rather than running independently.
- Emit a warning at threshold N but only suppress at threshold 2N, giving the model a chance to self-correct.

## Current behavior

Threshold is hardcoded at 5. Triggering it produces one suppressed turn, then resets.

## Request

Confirm the intended design rationale for 5 and decide whether to raise it, make it configurable, or replace it with a smarter heuristic.

### LINKED ISSUE #831: Revert non-standard local model workarounds from #823; document design stance
## Context

PR #823 added four fixes for Gemma 4 via LM Studio:

1. **Log deserialization failures** — `tracing::debug` on JSON parse failure ✅ keep
2. **`thinking_content` / `thought_content` fields** — non-standard field names ❌ revert
3. **`data:` without space in SSE** — spec-valid but added for LM Studio ❌ revert
4. **TUI thinking buffer flush at 120 chars** — UX fix for models with no newlines ❌ revert

## Problem

Fixes 2–4 add complexity to support local model servers (LM Studio) that diverge from
the OpenAI API format. This creates more problems than value:

- **Non-standard field names** (`thinking_content`, `thought_content`) are not documented
  in the OpenAI API spec. Supporting them means chasing every local server's naming whims.
- **SSE `data:` without space** is technically spec-valid but only seen from non-conforming
  servers. Standard OpenAI, Anthropic, and Google APIs all use `data: ` with a space.
- **120-char thinking flush** is a workaround for models that emit thinking without
  newlines — a model quality issue, not a koda issue.

Small local models are not koda's target use case. They produce lower quality output,
use non-standard APIs, and the engineering effort to support their quirks diverts from
higher-value work.

## Plan

1. Revert `thinking_content` and `thought_content` fields from `StreamDelta`
2. Revert `data:` (no space) SSE parsing — keep only `data: ` (with space)
3. Revert 120-char TUI thinking buffer flush
4. Keep the `tracing::debug` on parse failure (useful for all providers)
5. Remove associated tests
6. Add design principle to DESIGN.md: no over-optimization for small local models

## Design Principle

> Koda targets frontier models via standard APIs. We do not invest engineering
> effort in workarounds for small local models or non-conforming API servers.
> If a local server diverges from the OpenAI API spec, the fix belongs in the
> server, not in koda.

Traces to P1 (build only what we need) and P3 (build for the world six months from now —
local model quirks are temporary, frontier model capabilities compound).

lijunzh:
## Summary

Reverts Gemma 4-specific workarounds from #823 and replaces our over-engineered loop detection with Gemini CLI's approach, informed by a code review of Claude Code, Codex, and Gemini CLI.

Closes #831.
Closes #830 — the answer to "is 5 the right threshold for tool suppression?" is: the entire mechanism violates P3, so we deleted it.

## Key findings from competitor analysis

| Agent | Loop detection | Dedup | Per-turn cap | Tool-only suppression |
|---|---|---|---|---|
| **Claude Code** | ❌ None | ❌ | ❌ | ❌ |
| **Codex** | ❌ None | ❌ | ❌ | ❌ |
| **Gemini CLI** | ✅ Consecutive (5x) + LLM judge | ❌ | ❌ | ❌ |
| **Koda (before)** | Windowed fingerprint + saturation | ✅ | ✅ (20) | ✅ (5 turns) |
| **Koda (after)** | ✅ Consecutive (5x) + feedback | ❌ | ❌ | ❌ |

### Gemma 4 revert
- Remove `thinking_content` / `thought_content` fields from OpenAI-compat `StreamDelta`
- Remove partial-line thinking flush in TUI renderer
- Remove Gemma 4-specific tests in stream_collector
- Add design principle to DESIGN.md: standard APIs only, no per-model workarounds

### Loop detection refactor (Gemini CLI approach)

**Removed** — artificial safety mechanisms no other agent has:
- Tool call deduplication (was papering over Gemma 4 emitting 66 identical calls)
- Windowed fingerprint tracking (`REPEAT_THRESHOLD` in a sliding window)
- Tool-name saturation detection (`NAME_SATURATION_THRESHOLD`)
- Consecutive tool-only response suppression (`TOOL_ONLY_RESPONSE_LIMIT`) — closes #830
- Per-turn tool call cap (`MAX_TOOL_CALLS_PER_TURN = 20`)

**Added** — Gemini CLI's approach:
- Simple consecutive-identical-call detection (same tool+args 5x in a row)
- **Feedback injection** on first detection — injects "take a step back" message, continues
- **Hard stop** only on second detection — model ignored feedback
- Hard iteration cap (200) as absolute ceiling

### SSE spec + comment cleanup
- Restore `data:` (no space) SSE parsing — spec says space is optional
- Fix "local model" framing in comments across 6 files
- Document anti-over-compensation principle in DESIGN.md under P3

## What still works

| Feature | Status |
|---|---|
| LM Studio / Ollama / vLLM via OpenAI-compat API | ✅ |
| Context window auto-detect (LM Studio + Ollama) | ✅ |
| `<think>` tag filter (DeepSeek, Qwen) | ✅ |
| Tool name normalization (snake_case → PascalCase) | ✅ |
| Hard iteration cap (200) | ✅ |
| Consecutive loop detection + feedback injection | ✅ (new) |

## Net impact

- **-500+ lines** removed across the branch
- Simpler, more maintainable loop detection
- No more silent dedup hiding model quality issues
- Feedback injection gives models a chance to recover (vs hard stop)

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -114,6 +114,17 @@ limitations — they'll be obsolete before they're stable.
   workarounds for individual models' non-conforming behavior. If a model
   emits malformed output, the fix belongs upstream. See
   [#831](https://github.com/lijunzh/koda/issues/831) for rationale
+- **Don't over-compensate for weak models.** If a model emits 66 identical
+  tool calls, the answer is "use a better model," not a dedup layer that
+  silently papers over the problem. Safety mechanisms should match what
+  frontier agents actually ship — not what we imagine might go wrong.
+  A code review of Claude Code (zero loop detection), Codex (zero), and
+  Gemini CLI (consecutive-call detection + feedback injection) informed
+  our approach: consecutive identical calls → feedback injection → hard
+  stop only if the model ignores feedback. No windowed fingerprinting,
+  no tool-name saturation, no tool-only suppression, no per-turn caps.
+  See [#823](https://github.com/lijunzh/koda/issues/823) for the full
+  analysis
 - **Expand the surface.** Email, calendar, knowledge management — these
   are bets on where AI will be, not where it is
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -107,6 +107,13 @@ limitations — they'll be obsolete before they're stable.
   limitations are tomorrow's non-issues
 - **Don't scaffold around weakness.** Model tiers, capability probes, and
   verbose fallback prompts are building for today. Delete them
+- **Frontier models, standard APIs.** Koda targets frontier-class models
+  (Claude, GPT-4o, Gemini, o3/o4-mini) via their standard APIs. We do not
+  invest engineering effort in workarounds for small local models or
+  non-conforming API servers. If a local server diverges from the OpenAI
+  API spec, the fix belongs in the server, not in koda. Local model quirks
+  are temporary — frontier model capabilities compound. See
+  [#831](https://github.com/lijunzh/koda/issues/831) for rationale
 - **Expand the surface.** Email, calendar, knowledge management — these
   are bets on where AI will be, not where it is
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -108,11 +108,11 @@ limitations — they'll be obsolete before they're stable.
 - **Don't scaffold around weakness.** Model tiers, capability probes, and
   verbose fallback prompts are building for today. Delete them
 - **Frontier models, standard APIs.** Koda targets frontier-class models
-  (Claude, GPT-4o, Gemini, o3/o4-mini) via their standard APIs. We do not
-  invest engineering effort in workarounds for small local models or
-  non-conforming API servers. If a local server diverges from the OpenAI
-  API spec, the fix belongs in the server, not in koda. Local model quirks
-  are temporary — frontier model capabilities compound. See
+  via their standard APIs — whether cloud-hosted (Claude, GPT-4o, Gemini)
+  or locally served (Qwen, DeepSeek, Llama via LM Studio/Ollama/vLLM).
+  We support the OpenAI-compatible API spec faithfully but do not add
+  workarounds for individual models' non-conforming behavior. If a model
+  emits malformed output, the fix belongs upstream. See
   [#831](https://github.com/lijunzh/koda/issues/831) for rationale
 - **Expand the surface.** Email, calendar, knowledge management — these
   are bets on where AI will be, not where it is
--- a/koda-cli/src/tui_render.rs
+++ b/koda-cli/src/tui_render.rs
@@ -156,19 +156,6 @@ impl TuiRenderer {
                         ]),
                     );
                 }
-                // Flush partial line if the buffer is getting long (prevents
-                // the UI from appearing frozen when models like Gemma 4 emit
-                // long thinking stretches without newlines — issue #823).
-                if self.think_buf.len() > 120 {
-                    let partial = std::mem::take(&mut self.think_buf);
-                    tui_output::emit_line(
-                        buffer,
-                        Line::from(vec![
-                            Span::styled("  \u{2502} ", DIM),
-                            Span::styled(partial, DIM),
-                        ]),
-                    );
-                }
             }
             EngineEvent::ThinkingDone => {
                 if !self.think_buf.is_empty() {
--- a/koda-core/src/inference.rs
+++ b/koda-core/src/inference.rs
@@ -46,7 +46,7 @@ use crate::inference_helpers::{
     estimate_tokens, is_context_overflow_error, is_image_rejection_error, is_rate_limit_error,
     is_server_error, rate_limit_backoff,
 };
-use crate::loop_guard::LoopDetector;
+use crate::loop_guard::{LoopAction, LoopDetector};
 use crate::persistence::Persistence;
 use crate::providers::{
     ChatMessage, ImageData, LlmProvider, StreamChunk, TokenUsage, ToolCall, ToolDefinition,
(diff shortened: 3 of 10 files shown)


lijunzh:
## 🔍 Inference Loop Review: Local Model Over-Engineering

Thorough review of `koda-core/src/` for code that exists primarily to support small local models. Findings below, ranked by severity.

---

### 🔴 Should Remove (high confidence)

#### 1. `query_lmstudio_capabilities()` and `query_ollama_capabilities()` — `openai_compat.rs:509-580`
**72 lines** of LM-Studio-specific (`/api/v0/models`) and Ollama-specific (`/api/show`) API calls. Frontier providers (OpenAI, Anthropic, Gemini) have their own capability query endpoints. The `model_capabilities()` method tries LM Studio first, then Ollama — this ordering bakes in a local-model bias.

**Recommendation:** Remove both methods. The hardcoded `model_context.rs` lookup is the fallback and works fine. If users set `--base-url` to a custom endpoint, they can also set context window via config.

#### 2. `"LM Studio sends empty string content with tool calls; normalize to None"` — `openai_compat.rs:435`
One-line workaround for LM Studio sending `""` instead of `null`. Frontier APIs do not do this.

**Recommendation:** Remove the `.filter(|c| !c.is_empty())` — or at minimum, remove the LM Studio comment and treat it as generic defensive code.

#### 3. `"local"` alias in `model_alias.rs`
Hardcoded `"local"` → `LMStudio` auto-detect pathway with `needs_auto_detect()` sentinel. This is ~20 lines of special-case code.

**Recommendation:** Remove the `LOCAL_ALIAS` constant, the `"local"` branch in `resolve()`, and `needs_auto_detect()`. Users can still use `--provider lmstudio --model auto` explicitly.

#### 4. `KODA_ACCEPT_INVALID_CERTS` localhost bypass — `providers/mod.rs:228-240`
Self-signed cert bypass scoped to localhost. Only relevant for local model servers.

**Recommendation:** Remove entirely. Local servers should use proper certs or HTTP.

---

### 🟡 Borderline (keep with caveats)

#### 5. `tool_normalize.rs` — **462 lines**
Massive alias map (`list_files` → `List`, `read_file` → `Read`, etc.) plus dedup + cap logic. The module doc literally says *"especially smaller/local ones via OpenAI-compatible APIs"*. Frontier models (Claude, GPT-4o, Gemini) always use exact tool names as given.

**However:** The dedup + cap logic (`MAX_TOOL_CALLS_PER_TURN = 20`) is reasonable defensive code for any model. The fast path (canonical name → self-mapping) is O(1).

**Recommendation:** Keep the `normalize_tool_name()` function and dedup/cap, but **gut the alias map** — remove all snake_case/camelCase aliases (`list_files`, `read_file`, `run_command`, etc.). Keep only the self-mappings for canonical names. If a frontier model ever emits the wrong name, that is a model bug to report, not a koda workaround.

#### 6. `stream_tag_filter.rs` — **662 lines**
Handles `<think>`, `<thinking>`, `<reasoning>`, `<reflection>` tags, plus special tokens (`<|im_start|>`, `<|endoftext|>`, etc.).

**Mitigating factor:** DeepSeek (frontier-class) uses `<think>` tags via OpenRouter. The think-tag filter has legitimate frontier use. The special token stripping is local-model-only though.

**Recommendation:** Keep the think-tag filter. Consider removing the special token stripping (`<|im_start|>`, `<|endoftext|>`, etc.) — those are tokenizer artifacts from local models.

#### 7. `TOOL_ONLY_RESPONSE_LIMIT` + tool suppression — `inference.rs:700-718`
Suppresses tool definitions after 5 consecutive tool-only responses. Comment says *"prevents local models from looping infinitely."*

**Mitigating factor:** Any model can get stuck in a tool-calling loop. This is a reasonable safety valve.

**Recommendation:** Keep, but update the comment to remove the local-model framing.

---

### ✅ Fine as-is

- **`is_image_rejection_error()`** — matches error strings from multiple providers including OpenAI and Gemini. Defensive, useful for all providers.
- **`model_context.rs`** — hardcoded lookup for known models. `MIN_CONTEXT = 4096` safety floor is fine.
- **Provider table in `mod.rs`** — listing LM Studio/Ollama is documentation, not code complexity.
- **`sse_handle.abort()` on Ctrl+C** — comment mentions LM Studio but the abort is useful for any SSE stream.

---

### Summary

| Category | Items | Est. lines to remove |
|---|---|---|
| 🔴 Remove | LM Studio/Ollama cap queries, empty-content workaround, `"local"` alias, TLS bypass | ~120 |
| 🟡 Gut alias map | `tool_normalize.rs` aliases (keep dedup/cap) | ~80 |
| 🟡 Strip special tokens | `stream_tag_filter.rs` SPECIAL_TOKENS | ~30 |
| 🟡 Fix comments | Remove local-model framing from loop guard, sse abort | ~5 |

Total: ~235 lines of local-model-specific code to remove/simplify across the codebase.

PR #880, opened 2026-04-13
fix(inference): sentinel system msg anchors model to correct task after network-error resume (#875, #877)

### LINKED ISSUE #875: bug: "continue" after unexpected inference termination re-sends full irrelevant session history
## Summary

When an inference pipeline terminates unexpectedly mid-task and the user types `continue` to resume, Koda sends the entire session history from the beginning — including early, unrelated messages — rather than just the context relevant to the interrupted task.

## Observed Behavior

In a session where the first two turns were unrelated `ls` commands followed by a long brainstorming conversation, the inference loop crashed during source-code exploration. The user typed `continue`. Koda sent the **full history from the very beginning** (including the `ls` outputs) to the model, which:

1. Wastes context window tokens on irrelevant early messages
2. Confuses the model about what it was doing (early commands create misleading context)
3. The context percentage shown in the status bar confirmed an unexpectedly large history was being sent

## Root Cause

`load_context()` in `koda-core/src/db/queries.rs:293` loads **all** messages for a session from SQLite. There is no mechanism to:

- Mark a turn as "interrupted" and resume from that boundary
- Exclude early off-topic messages that pre-date the interrupted task
- Compact irrelevant preamble before sending context to the model

When inference fails mid-turn, the partial turn state (tool calls in progress, intermediate reads) may or may not be committed to the DB. On `continue`, the model sees everything back to session start.

## Impact

In a session with unrelated warm-up commands followed by a long task, the effective context sent to the model can be 2–3× larger than necessary. Combined with the context % underreporting bug (#874), users have no visibility into this.

## Possible Improvements

1. **Turn interruption marker**: When a turn exits with an error, write a sentinel message to the DB (e.g., `system: "Previous turn terminated unexpectedly"`) so the model immediately knows what happened on `continue`.

2. **Microcompact early irrelevant messages**: After an unexpected termination, trigger `microcompact.rs` on early messages that are clearly off-topic relative to the most recent N messages. The model can determine relevance.

3. **Session task context**: Track a "current task" pointer in the DB so `load_context` can prioritize recent task-relevant messages over session preamble.

4. **Surface the issue to users**: At minimum, show users how many messages are being sent and from when, so they can manually `/compact` before typing `continue`.

## Relevant Files

- `koda-core/src/db/queries.rs:293` — `load_context` loads all session messages
- `koda-core/src/inference.rs:126` — `load_context` call site
- `koda-core/src/microcompact.rs` — existing lightweight compaction (could be triggered here)
- `koda-cli/src/tui_handlers_inference.rs:149-160` — turn error handling (shows error but takes no recovery action)

### LINKED ISSUE #877: bug: after pipeline termination, model responds to stale history messages instead of the user's "continue"
## Summary

When the inference pipeline terminates unexpectedly mid-task and the user types `continue`, the model sometimes responds to an *earlier* message from the session history rather than the actual `continue` request — producing a completely wrong response.

## Observed Behavior

Session flow:
1. Session opened with two `ls` commands (early, unrelated warm-up)
2. Long brainstorming conversation about issue #851
3. Source-code research across many tool calls (Bash reads of Claude Code source)
4. Inference pipeline terminates unexpectedly mid-research
5. User types: `continue`

**Expected:** Model resumes the source-code research and summarizes its findings.

**Actual:** Model's internal thinking reads *"The user just typed `ls` twice (likely testing the command). Let me list the files in the current directory."* and responds with a Koda workspace overview — as if the session had just started.

## Root Cause

This is a compounding failure of the same root cause as #875:

1. `load_context()` loads the entire session history on resume, including the two unrelated `ls` commands from session start.
2. The model receives the `continue` message surrounded by a very large context window full of tool results and earlier exchanges.
3. After many inference iterations with tool calls in between, the model loses track of which message is the *current* user request and latches onto an earlier user message (`ls`) that appears "clean" relative to the surrounding noise.

The result is a **correctness failure**, not just a context efficiency problem: the model produces a response to the wrong user message. #875 covers the efficiency dimension (unnecessary history); this issue covers the correctness dimension.

## Why This Is Distinct From #875

| | #875 | This issue |
|---|---|---|
| Problem | Irrelevant history wastes context tokens | Model responds to the *wrong* user message |
| Symptom | High context usage, slow inference | Completely incorrect response |
| Fix direction | Trim/bound history on resume | Ensure the current user message is unambiguously identifiable after many tool iterations |

## Potential Fixes

1. **Sentinel boundary (shared with #875):** A `system: "Previous turn terminated unexpectedly. Current request: {user_message}"` message injected at resume time keeps the current intent visible after many tool-call iterations.

2. **Re-anchor at turn boundaries:** When the inference loop completes a full iteration (after all tool results), re-emit the original user message as a reminder so the model doesn't lose track in long multi-step flows.

3. **Last-message position guarantee:** Ensure the most recent user message is *always* the last non-tool message in the assembled context, even after compaction/microcompaction shuffles earlier messages.

## Relevant Files

- `koda-core/src/inference.rs:126` — `load_context` call, history assembly
- `koda-core/src/inference_helpers.rs:70` — `assemble_messages` (order of messages)
- `koda-cli/src/tui_handlers_inference.rs:149-160` — turn error handling (no recovery action taken)

lijunzh:
## Problem (shared root cause)

`load_context()` returns the full unfiltered session history on every resume. After an unexpected network-error termination:

- **#877 (correctness):** The model receives `continue` surrounded by a large history, loses track of which message is the current request, and latches onto an earlier unrelated user message (e.g., `ls`)—producing a completely wrong response.
- **#875 (efficiency):** The bloated context includes all session history from the beginning, wasting 2–3× tokens on irrelevant preamble.

### Fix 1 — `inference.rs`: write a `[System]` sentinel to DB on network drop

When `stream_result.network_error.is_some()`, before returning, insert a system message that re-anchors the model:

```
[System] The previous inference turn terminated unexpectedly due to a network
error. The session history above is intact. The user's pending request was:
"{last_user_message}". Please resume from where you left off.
```

The sentinel is the last message `load_context()` returns on the next turn, so the model sees an unambiguous "you are here" marker immediately before the user's "continue". Full history is still present (deeper compaction is tracked separately), but the anchor prevents the wrong-message correctness failure.

Write is best-effort: DB failure is warn-logged but never propagates.

### Fix 2 — `tui_handlers_inference.rs`: add /compact hint to the error banner

After the red `✗ Turn failed` line, add a yellow hint:

> Tip: run /compact to trim unrelated early messages before typing "continue".

This directly surfaces the #875 mitigation at the exact moment users need it.

## Tests

Added to `koda-core/tests/inference_recovery_test.rs`:
- `network_error_writes_sentinel_with_user_request` — verifies sentinel is in DB and quotes the pending request
- `network_error_writes_generic_sentinel_when_no_user_message` — edge case: empty user message still writes sentinel

All 6 recovery tests pass. ✅

Closes #875
Closes #877

### DIFF
--- a/koda-cli/src/tui_handlers_inference.rs
+++ b/koda-cli/src/tui_handlers_inference.rs
@@ -155,6 +155,19 @@ impl TuiContext {
                                     ),
                                 ]),
                             );
+                            // #875: guide the user before they type "continue"
+                            // — irrelevant early history is the most common cause
+                            // of the model responding to the wrong message (#877).
+                            self.scroll_buffer.push(
+                                Line::from(vec![
+                                    Span::raw("  "),
+                                    Span::styled(
+                                        "Tip: run /compact to trim unrelated early \
+                                         messages before typing \"continue\".",
+                                        Style::default().fg(Color::Yellow),
+                                    ),
+                                ]),
+                            );
                         }
                         break;
                     }
--- a/koda-core/src/inference.rs
+++ b/koda-core/src/inference.rs
@@ -839,6 +839,30 @@ pub async fn inference_loop(ctx: InferenceContext<'_>) -> Result<()> {
         // Discard the partial response — storing it would corrupt the session.
         if stream_result.network_error.is_some() {
             sse_handle.abort();
+            // Write a sentinel system message so the model re-anchors to the
+            // correct task on the next "continue" (#875, #877).  Without this
+            // the model loads full session history and can latch onto an early
+            // unrelated user message instead of the current pending request.
+            let last_user = db.last_user_message(session_id).await.unwrap_or_default();
+            let sentinel = if last_user.is_empty() {
+                "[System] The previous inference turn terminated unexpectedly \
+                 due to a network error.  The session history above is intact. \
+                 Please resume from where you left off."
+                    .to_string()
+            } else {
+                format!(
+                    "[System] The previous inference turn terminated unexpectedly \
+                     due to a network error.  The session history above is intact. \
+                     The user's pending request was: \"{last_user}\". \
+                     Please resume from where you left off."
+                )
+            };
+            if let Err(e) = db
+                .insert_message(session_id, &Role::System, Some(&sentinel), None, None, None)
+                .await
+            {
+                tracing::warn!("Failed to write interruption sentinel to DB: {e:#}");
+            }
             return Ok(());
         }
 
--- a/koda-core/tests/inference_recovery_test.rs
+++ b/koda-core/tests/inference_recovery_test.rs
@@ -148,3 +148,76 @@ async fn test_context_overflow_too_few_messages_fails() {
         "error should mention overflow: {err}"
     );
 }
+
+// ── Sentinel message tests (#875, #877) ─────────────────────
+
+/// After a network-error drop, a system sentinel must be written to the DB
+/// so the model can re-anchor to the correct task on the next "continue".
+/// Without the sentinel, load_context() returns the full unfiltered history
+/// and the model latches onto an earlier unrelated user message (#877).
+#[tokio::test]
+async fn network_error_writes_sentinel_with_user_request() {
+    let env = Env::builder().max_context_tokens(100_000).build().await;
+    env.insert_user_message("research the inference loop implementation").await;
+
+    let provider = MockProvider::new(vec![MockResponse::NetworkError {
+        partial_text: "I'll start by".into(),
+        error: "connection reset by peer".into(),
+    }]);
+
+    let (result, _events) = env.run_inference_result(&provider).await;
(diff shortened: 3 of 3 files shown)


lijunzh:
Superseded by #885.

#885 fixes the root cause: `load_context()` now filters incomplete assistant messages (`completed_at IS NULL`) at the SQL level, so after any interruption the model gets a clean context ending with the user's pending request — no sentinel needed.

The sentinel was compensating for a leaky `load_context()`. With #885 merged, injecting it would just consume context tokens and add noise. The two PRs' tests actively conflict: #880 asserts the sentinel IS present, #885 asserts it is NOT.

The `/compact` tip UX from #880 is worth keeping; tracked separately.

PR #1055, opened 2026-04-26
fix(#1054): validate sub-agent session_id before INSERT to avoid raw FK error

### LINKED ISSUE #1054: Sub-agent invocation fails with FOREIGN KEY constraint when session_id is unknown
## Summary

When the model invokes `InvokeAgent` (especially several in parallel) and passes a `session_id` argument that doesn't exist in the `sessions` table, the dispatch fails with a SQLite FOREIGN KEY error. The user-visible message is:

```
Error invoking sub-agent: error returned from database: (code: 787) FOREIGN KEY constraint failed
```

Reproduced organically with the `minimax-m2.7` model (openai-compat provider) when asked to start a few sub-agents to explore the home directory. All three parallel `InvokeAgent` calls failed with the same error. Full transcript excerpt below.

## Likely cause

In `koda-core/src/sub_agent_dispatch.rs` around line 515:

```rust
let sub_session = match session_id {
    Some(id) => id,                     // <-- used as-is, never validated
    None => {
        let sid = db.create_session(...).await?;
        ...
        sid
    }
};

db.insert_message(&sub_session, &Role::User, Some(prompt), None, None, None).await?;
```

The `session_id` arg is documented in `koda-core/src/tools/agent.rs:80` as *"Optional session ID to continue a previous sub-agent conversation"*. When the model hallucinates an ID, copies one from another context, or supplies an ID that has been purged, `insert_message` violates the FK to `sessions(id)` (defined in `koda-core/src/db/mod.rs:167`) and the whole InvokeAgent fails with an opaque SQLite error.

This is easy for some models to trip into when fanning out in parallel — they invent UUIDs that look plausible.

## Suggested fix

When `session_id` is `Some(id)`:

1. Verify the session row exists (and ideally that its `agent_name` matches), then proceed.
2. If it doesn't, return a clear `[ERROR: sub-agent ...]` message — matching the structural-failure contract documented on the InvokeAgent tool — so the model can re-strategize, instead of surfacing the raw SQLite FK error.

Secondary tweak: tighten the `session_id` description in the InvokeAgent schema so the model knows it must only pass an ID returned from a prior `InvokeAgent` call in the same conversation, never a guess.

## Reproduction context

- Koda session: `267eecb7-8d3c-4536-a50b-5407753db0d0`
- Model: `minimax-m2.7`
- Provider: `openai-compat`
- Project: `/Users/lijun`
- Date: 2026-04-26

## Transcript excerpt

User asked: *"start a few sub agents to explore the home directory and provide recommendation to improve it."*

The assistant fired three parallel `InvokeAgent` tool calls. All three returned:

```
Error invoking sub-agent: error returned from database: (code: 787) FOREIGN KEY constraint failed
Error invoking sub-agent: error returned from database: (code: 787) FOREIGN KEY constraint failed
Error invoking sub-agent: error returned from database: (code: 787) FOREIGN KEY constraint failed
```

The assistant then fell back to running the exploration manually with `Bash`/`List`, which worked.

lijunzh:
Fixes #1054.

## Problem

When the model invoked `InvokeAgent` with a `session_id` argument that didn't exist in the `sessions` table — most often a hallucinated UUID during parallel fan-out, but also a copy from another context or an ID that had been purged — the dispatch path passed the ID straight into `Persistence::insert_message`. The `messages.session_id → sessions.id` FK then raised SQLite error 787 and the user-facing tool result was:

```
Error invoking sub-agent: error returned from database: (code: 787) FOREIGN KEY constraint failed
```

The model has no recovery strategy for that string. Reproduction with `minimax-m2.7` fanning out three parallel `InvokeAgent` calls failed three-for-three.

## Fix

1. **Add `Persistence::session_exists`** — a cheap `SELECT 1 FROM sessions WHERE id = ? LIMIT 1` (single index probe, no row materialisation). Hot path: every `InvokeAgent` call that supplies a `session_id` runs this once before the FK-bearing INSERT.

2. **Validate `session_id` in `execute_sub_agent`** — at the top of the `Some(id)` branch in `koda-core/src/sub_agent_dispatch.rs`. On miss, return the standard `[ERROR: sub-agent ...]` marker (cached, same as the other structural-failure markers) so the model treats it as metadata and re-strategizes — same contract `iteration_cap_marker` and `workspace_provision_failure_marker` already use. The marker explicitly tells the model to omit `session_id` to start a fresh session, and warns against inventing IDs on retry.

3. **Tighten the `InvokeAgent` schema** — the `session_id` description now says it must only be an ID returned by a prior `InvokeAgent` call in the same conversation, never invented or copied, and that omitting the argument is the right move for a fresh sub-agent session.

## Tests

- `db::tests::test_session_exists` — pins the three behaviours we depend on:
  - fresh session found,
  - unknown UUID rejected,
  - deleted session rejected (the post-purge case from the issue).
- `sub_agent_dispatch::issue_1054_tests` — six contract tests for the new marker (`[ERROR:` prefix, agent name, bad id, omit-hint, no-invent warning, single-line) following the same pattern as the existing `b18_tests` (iteration cap) and `b21_tests` (workspace provision).

```
running 7 tests
test sub_agent_dispatch::issue_1054_tests::marker_has_error_prefix ... ok
test sub_agent_dispatch::issue_1054_tests::marker_warns_against_inventing_ids ... ok
test sub_agent_dispatch::issue_1054_tests::marker_includes_agent_name ... ok
test sub_agent_dispatch::issue_1054_tests::marker_is_single_line ... ok
test sub_agent_dispatch::issue_1054_tests::marker_includes_bad_session_id ... ok
test sub_agent_dispatch::issue_1054_tests::marker_tells_model_to_omit_session_id ... ok
test db::tests::test_session_exists ... ok

test result: ok. 7 passed; 0 failed
```

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the full `cargo test --workspace --features koda-core/test-support` all pass on this branch (modulo two pre-existing `koda-cli/src/transcript.rs` failures that also fail on `main` and are unrelated to this change).

## Out of scope

End-to-end coverage of the dispatch short-circuit (validate → cache → return) requires a mock `Persistence` injection point that doesn't exist on the public seam yet. The unit-level coverage above plus the manual repro from the issue is the regression net for now.

### DIFF
--- a/koda-core/src/db/queries.rs
+++ b/koda-core/src/db/queries.rs
@@ -171,6 +171,19 @@ impl Persistence for Database {
         Ok(id)
     }
 
+    /// #1054: cheap existence check for sub-agent dispatch validation.
+    ///
+    /// `SELECT 1 ... LIMIT 1` over the primary key — a single index probe,
+    /// no row materialisation. Hot path: every `InvokeAgent` call that
+    /// passes a `session_id` runs this once before the FK-bearing INSERT.
+    async fn session_exists(&self, session_id: &str) -> Result<bool> {
+        let row: Option<(i64,)> = sqlx::query_as("SELECT 1 FROM sessions WHERE id = ? LIMIT 1")
+            .bind(session_id)
+            .fetch_optional(&self.pool)
+            .await?;
+        Ok(row.is_some())
+    }
+
     /// Insert a message into the conversation log.
     async fn insert_message(
         &self,
--- a/koda-core/src/db/tests.rs
+++ b/koda-core/src/db/tests.rs
@@ -43,6 +43,45 @@ async fn test_create_session() {
     assert!(!id.is_empty());
 }
 
+/// **#1054**: `session_exists` is the cheap pre-INSERT validator the
+/// sub-agent dispatch path uses to reject hallucinated `session_id`
+/// arguments before they trip the `messages.session_id → sessions.id`
+/// FK and surface SQLite error 787 to the model.
+///
+/// Pin the three behaviours we depend on:
+///
+/// 1. Returns `true` for a freshly-created session.
+/// 2. Returns `false` for a syntactically-valid-but-unknown UUID.
+/// 3. Returns `false` for a deleted session (covers the "session was
+///    purged between turns" case from the issue).
+#[tokio::test]
+async fn test_session_exists() {
+    let (db, _tmp) = setup().await;
+
+    // 1. Fresh session is found.
+    let id = db.create_session("default", _tmp.path()).await.unwrap();
+    assert!(
+        db.session_exists(&id).await.unwrap(),
+        "freshly-created session must be found"
+    );
+
+    // 2. Unknown UUID is rejected (the hallucination case).
+    assert!(
+        !db.session_exists("deadbeef-0000-0000-0000-000000000000")
+            .await
+            .unwrap(),
+        "unknown session_id must return false (the FK-error case)"
+    );
+
+    // 3. Deleted session is rejected (the purge case).
+    let deleted = db.create_session("default", _tmp.path()).await.unwrap();
+    assert!(db.delete_session(&deleted).await.unwrap());
+    assert!(
+        !db.session_exists(&deleted).await.unwrap(),
+        "deleted session_id must return false"
+    );
+}
+
 #[tokio::test]
 async fn test_insert_and_load_messages() {
     let (db, _tmp) = setup().await;
--- a/koda-core/src/persistence.rs
+++ b/koda-core/src/persistence.rs
@@ -192,6 +192,15 @@ pub trait Persistence: Send + Sync {
 
     /// Create a new session, returning its unique ID.
     async fn create_session(&self, agent_name: &str, project_root: &Path) -> Result<String>;
+    /// Returns `true` if a session row with the given ID exists.
+    ///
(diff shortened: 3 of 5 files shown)


lijunzh:
Closing in favour of a proper fix.

After surveying how Claude Code, Codex, Gemini CLI, and Zed handle sub-agent session identity, the validation layer added here is a bandaid over the wrong thing.

**Root cause**: `InvokeAgent` returns plain text. The model is never shown a session UUID. Therefore the only way a model ever populates `session_id` is by hallucinating one — and we wrote validation to catch the hallucination. That is not a fix; it's compensating for a broken API contract.

**What the references do instead:**
- **Codex** — the only system that lets the model reference an agent by ID — does it correctly: `spawn_agent` has a structured `output_schema` returning `{ agent_id, nickname }`. The model reads that ID from the tool result; it can never invent one. `resume_agent`/`wait_agent`/`send_input` then take that same system-issued ID.
- **Claude Code / Gemini CLI / Zed** — don't expose session IDs to the model at all. Session state is managed entirely host-side.

**Koda's `session_id` parameter is a dead parameter**: nothing ever returns a sub-agent session UUID to the model as tool output, so the parameter can never be legitimately populated. The correct fix is to remove it (YAGNI), not validate against its misuse.

A follow-up PR will delete `session_id` from the `InvokeAgent` schema and clean up the dispatch path.

Issue #1072, opened 2026-04-27
sandbox: code review findings vs. CC / Codex / Gemini — three concrete gaps + design wins


lijunzh:
## Sandbox Code Review — Peer Comparison (#934 follow-up)

> Covering `koda-sandbox/` + `koda-core/src/sandbox.rs` against Claude Code (CC), Codex, Gemini CLI, and Zed. All phase references map to the #934 roadmap.

---

## TL;DR

Koda's sandbox is the most principled of the five implementations in terms of formal correctness, design documentation, and security-invariant coverage. There are **three concrete gaps** worth closing before #934 is declared done, and several design wins worth preserving and protecting from future regressions.

---

### 1. Policy Composition Lattice (`SandboxPolicy::compose`)

The formal lattice semantics are unique among the peers:

| Field | Composition rule |
|---|---|
| `deny_read` / `deny_write` | Union — accumulate, never drop |
| `allow_write` | Intersect — narrow, never widen |
| `limits` (wall/cpu/mem/fds) | Minimum — strictest wins |
| `trust` | Strictest |

None of the peers have this:
- **CC** rebuilds from a Zod config object per invocation — no composition guarantee
- **Codex** derives child config from user-land config files — a malicious `.koda.toml` could widen child policy
- **Gemini** passes a `SandboxContext` struct through but has no formal composition rule

The mechanical test that enforces "policy must come from `policy_for_agent`, never from runtime config" catches regressions automatically. **Do not remove it.**

### 2. Path Defense Depth

Symlink chain tracing, dangling anchor guards, FIFO/socket rejection, and `effective_chain_depth` accounting go meaningfully deeper than CC's `realpath` call and Gemini's `resolveHome` helper. Each attack variant has a named, pinned test.

### 3. Egress Proxy — RFC 6125 Wildcard Matching

RFC 6125 single-label semantics are stricter than peers. CC's allowlist uses simple `endsWith` — `*.npmjs.org` in CC's implementation would match `evil.npmjs.org.attacker.com`. Koda's prefix/suffix decomposition with the no-dot check closes that phishing vector.

### 4. Violation Ring Buffer + Per-Kind Tallies

`SandboxViolationStore` with bounded ring, drain-on-read, `count_by_kind` for Phase-5 telemetry, and `global_store()` singleton. The "clones share the inner buffer" test pins the multi-slot behavior that would silently break otherwise.

### 5. Design Documentation Quality

[...]

PR #1073, opened 2026-04-27
sandbox: close #1045 – Gap 1 + Gap 3 from #1072 spec

### LINKED ISSUE #1045: Code review: DESIGN.md drift — dead code, speculative protocol surface, and the repl/tui_commands two-place dispatch
# Code Review: Drift between DESIGN.md and the implementation

## Summary

Cross-referenced the koda codebase against `DESIGN.md` and against four reference projects (Claude Code, Codex, Gemini CLI, Zed). The architecture genuinely matches the doc — engine/CLI split is real, sub-agent invariants are tested, the protocol boundary is `Serialize`-clean. **No structural problems.** What follows is a list of concrete drift points where code or docs have crept ahead/behind the other and small refactors would re-align the two.

For perspective on file-size discipline:

| Project | Largest single file | Architecture doc? |
|---|---|---|
| **koda** (`koda-core/src/inference.rs`) | 1,545 LOC | ✅ `DESIGN.md` |
| codex (`core/src/session/mod.rs`) | 3,328 LOC (+7,859 test) | ⚠️ AGENTS.md is process, not design |
| zed (`crates/agent/src/thread.rs`) | 4,456 LOC | ⚠️ AGENTS.md, no architecture doc |
| gemini-cli (`packages/core/src/config/config.ts`) | 3,903 LOC | ⚠️ ROADMAP only |

Koda is leaner than every reference and uniquely the only one with a published "why" doc. The findings below are nits at this level — none require touching core logic.

---

## 🟥 P1 violations — "delete aggressively"

These directly contradict P1 in `DESIGN.md`:

> "Build only what we need… Features that were built but aren't used should be deleted — git preserves history. Carrying dead code forward degrades every future decision."

### Finding 1 — `koda-core/src/approval.rs` is self-confessed dead code

```rust
//! The approval *mode* logic has moved to [`crate::trust`].
//! This module is kept only because `approval_flow.rs` references it —
//! it will be removed in a follow-up cleanup.
```

The "follow-up cleanup" never happened. The file is just three `pub use` lines re-exporting from `crate::trust`. There are exactly **3 call sites** (`koda-core/src/inference.rs:468-470`); both `tool_dispatch.rs` and `sub_agent_dispatch.rs` already use `crate::trust::*` directly.

**Fix**: delete `approval.rs`; replace `crate::approval::` with `crate::trust::` in `inference.rs`. ~10 minutes, ~3 lines net.

---

### Finding 2 — Speculative wire-protocol surface in `engine/event.rs`

`koda-core/src/engine/event.rs` carries multiple `#[allow(dead_code)]` markers on protocol types with comments like:

> "Currently handled client-side. Defined for wire protocol completeness."
> "Not yet consumed outside the engine module — wired in v0.2.0 server mode."

Affected:

- `EngineCommand::SlashCommand(SlashCommand)` and the entire `SlashCommand` enum (`Compact | SwitchModel | SwitchProvider | ListSessions | DeleteSession | SetTrust | Cost | Memory | Help | InjectPrompt`)
- `EngineCommand::Quit`
- `EngineCommand::UserPrompt { text, images }` (no engine consumer today)
- `ImageAttachment`

This is exactly the "extensibility for later" P1 forbids — every ACP client now has to think about variants will ever send.

**Fix**: delete them. When v0.2.0 server mode lands, *that PR* adds the variants it needs. Saves ~80 LOC of public protocol surface.

---

### Finding 3 — `repl::handle_command` claims to be shared with headless; it isn't

`koda-cli/src/tui_commands.rs:4` says:

> "Most commands are handled via `crate::repl::handle_command` (shared with headless)"

`koda-cli/src/headless.rs` does not call `repl::handle_command` anywhere. Headless mode has no slash commands by design — it runs one prompt and exits.

So the `ReplAction` enum + `handle_command` parser exists to let exactly **one** caller (`tui_commands.rs:53`) parse a slash-command string into a typed enum, which is *immediately* matched on by `tui_commands.rs::handle_slash_command` and dispatched to a handler in `tui_wizards.rs`.

This is a **two-place dispatch table**. Adding a new slash command needs:

1. Add a `ReplAction` variant in `repl.rs`
2. Add the parser arm in `repl.rs::handle_command`
3. Add the handler arm in `tui_commands.rs::handle_slash_command`
4. Often add a function in `tui_wizards.rs`

Steps 1+2 vs 3+4 is exactly the "engine match + frontend match" anti-pattern the protocol design (`engine/event.rs`) was meant to avoid — but `ReplAction` *isn't* part of the engine protocol, it's just internal CLI ceremony.

**Fix**: collapse the `repl.rs` parser arms directly into `tui_commands.rs::handle_slash_command`, delete `ReplAction`. The `PROVIDERS` constant moves to `widgets/provider_menu.rs` (its only real consumer outside `repl.rs`). Net delete: ~600 LOC including the parser test scaffold; per-command tests can move to `tui_commands.rs` or exercise the integration path.

This matches what zed/codex do — the slash-command parser lives where the slash-command is dispatched.

**Other call sites of `crate::repl::*`** to clean up:
- `koda-cli/src/tui_context/mod.rs:259` → uses `crate::repl::PROVIDERS`
- `koda-cli/src/tui_context/menus.rs:87,106` → uses `crate::repl::PROVIDERS`

---

### Finding 4 — `koda-cli` modules: doc says `pub`, code says `pub(crate)`

DESIGN.md L584-587:

> "**koda-cli modules must be `pub`** even though they have no external consumers. This is a deliberate visibility trade for documentation coverage — the guide agent needs to read slash command docs, TUI keybindings, and onboarding flow."

`koda-cli/src/lib.rs`: **41 of 41 modules are `pub(crate)`**. Zero are `pub`.

Consequence: the "docs.rs is our docs site" strategy is half-broken. `koda-core` modules render on docs.rs, `koda-cli` modules are completely invisible. The guide agent literally cannot fetch slash command / TUI keybinding docs as the design describes.

**Fix**: pick one.

- **Option A**: flip all 41 modules to `pub`. Matches DESIGN.md verbatim. Risk: docs.rs becomes noisy with internals.
- **Option B**: update DESIGN.md to "koda-cli is a binary; its API surface is the user-facing CLI plus `pub async fn run()`. User-facing docs live in `koda-core` rustdoc + the README."

Recommendation: **Option B**, since the binary really doesn't have an API surface and most of the listed "docs" (slash commands, keybindings) are user docs that belong in a guide, not rustdoc.

---

### Finding 5 — Sub-agent / dispatch file naming doesn't tell the story

DESIGN.md treats sub-agent lifecycle as one mental model under "Sub-Agent Lifecycle (P2, P3)". The implementation splits it across four files whose names don't signal their relationships:

| File | LOC | Actually contains |
|---|---:|---|
| `koda-core/src/tool_dispatch.rs` | 1,042 | Sequential / parallel / split-batch tool execution |
| `koda-core/src/sub_agent_dispatch.rs` | 1,464 | One sub-agent invocation (with B1–B21 invariants) |
| `koda-core/src/bg_agent.rs` | 1,442 | Background sub-agent registry & lifecycle |
| `koda-core/src/tools/agent.rs` | ~270 | `InvokeAgent` / `ListAgents` LLM tool definitions |

The split is defensible (cohesion within each file is high — the comment density on B1–B21 invariants in `sub_agent_dispatch.rs` would be lost if split further). The names just don't reflect the unified concept.

**Fix (optional)**: rename for clarity, or add `koda-core/src/sub_agent/mod.rs` that re-exports the three sub-agent files into a documented namespace. Cosmetic; improves on-ramp time.

Suggested renames:

- `bg_agent.rs` → `sub_agent_bg_registry.rs`
- `sub_agent_dispatch.rs` → `sub_agent_invocation.rs`
- `tool_dispatch.rs` → `tool_batch_executor.rs`

---

### Finding 6 — `bg_process.rs` vs `bg_task_tools.rs` split obscures the unification

In `koda-core/src/tools/`:

- `bg_process.rs` (672 LOC) — `BgRegistry` for background **shell processes**
- `bg_task_tools.rs` (781 LOC) — `ListBackgroundTasks` / `CancelTask` / `WaitTask` LLM tools that span both shell processes *and* sub-agents

Per DESIGN.md #996 (background-task management surface):

> "single unified concept 'background task' with two flavors distinguished by a prefixed id (`agent:N` / `process:N`)"

The user-facing unification is clean. The *internal* registries are split across `bg_agent.rs` (sub-agents) and `tools/bg_process.rs` (shells), with `bg_task_tools.rs` as the bridge. File names should signal the unification.

**Fix**: `bg_task_tools.rs` → `bg_tasks.rs` (or similar) and add a one-paragraph map in the module doc explaining the two-registry bridge. Reduces cognitive load when adding a third bg-task type someday.

---

## 🟩 Things that are right and shouldn't change

For the record, so future drift PRs don't accidentally undo these:

### File-size discipline matches P2's "cohesion over abstraction"

DESIGN.md P2 explicitly defends large cohesive files:

> "An 800-line file with one cohesive flow beats two 400-line files that require cross-file context-switching."

This is honored in the right places:

- `koda-core/src/inference.rs` (1,545 LOC) — one cohesive `inference_loop()` plus tightly-coupled helpers (`assemble_context`, `collect_stream`, `try_overflow_recovery`, `try_with_rate_limit`). Splitting forces redundant args on every helper. ✅
- `koda-core/src/tools/mod.rs` (~1,500 LOC) — one giant `match` in `ToolRegistry::execute()`; the design *requires* this for compile-time exhaustiveness (P2 "Tool Dispatch: Match Statement"). ✅
- `koda-core/src/sub_agent_dispatch.rs` (1,464 LOC) — one `execute_sub_agent` with the B1–B21 invariant comment density. Splitting risks losing invariant locality. ✅

For comparison: codex's `session/mod.rs` is 3,328 LOC, zed's `thread.rs` is 4,456 LOC, gemini-cli's `config.ts` is 3,903 LOC. Koda is downright restrained.

### Trait discipline matches P1/P2

Only ~8 traits in the entire workspace, every one justified by an actual second implementation or an explicit "swap backend later" call-out in DESIGN.md:

- `Persistence` (P3 — explicitly called out as "swap backend later, mock for tests")
- `LlmProvider` (multi-provider is the *core* P1 differentiator)
- `EngineSink` (engine-as-library boundary; CLI/server/null/buffering impls)
- `WorkspaceProvider` / `SandboxRuntime` / `FileSystem` (per-OS sandbox abstractions; multiple impls exist)
- `DropdownItem` (one widget, multiple item types — local type erasure, not a plugin point)
- `ChunkParser` (provider stream-collector helper)

Compare: zed has a `Tool` trait + dynamic dispatch + `Box<dyn Tool>` registry. Koda's discipline here is the cleanest in the comparison set.

### Sub-agent invariants are pinned by tests

The B1–B21 invariants in DESIGN.md are pinned by tests in `koda-cli/tests/regression_test.rs` and `koda-core/tests/guarantee_matrix_test.rs`. The doc names specific tests (`test_reject_auto_wire_tag_is_snake_case`, `test_can_parallelize_delete_owned_file_uses_tracker`) and they exist. ✅

### `clippy::too_many_arguments` allowances are pragmatic, not lazy

19 sites use `#[allow(clippy::too_many_arguments)]`. The price of explicit dependency injection in a no-globals engine. The `InferenceContext` struct exists for the cases where it pays for itself; one-off internal helpers stay positional. Pragmatic.

---

## Suggested action plan

In rough priority order (small/safe first):

- [ ] **Delete `koda-core/src/approval.rs`** — 10 min. (Finding 1)
- [ ] **Delete dead `EngineCommand` variants and `ImageAttachment`** — 30 min. (Finding 2)
- [ ] **Resolve the `pub` vs `pub(crate)` contradiction** — 5 min decision, then either flip or update DESIGN.md. (Finding 4)
- [ ] **Inline `repl::handle_command` into `tui_commands.rs`, delete `ReplAction`** — 1–2 hours. Largest LOC win (~600 lines net delete) and biggest workflow improvement for adding slash commands. (Finding 3)
- [ ] **Optional: rename pass for the four sub-agent-related files / merge `bg_*` naming** — half day. Cosmetic. (Findings 5, 6)

None are urgent. None require touching the actual logic. They're all "follow your own published rules a bit harder."

---

## What koda already does better than the references

So this isn't all nits — three things to keep doing:

1. **DESIGN.md as a living engineering log.** Every "we considered X and rejected because Y" is gold. Codex's AGENTS.md is process; zed's AGENTS.md is style; gemini-cli has a roadmap. None of them tell a contributor *why the architecture is shaped this way*.

2. **Four-mode sub-agent model with a single mental model.** Codex needs `spawn_agent` / `send_input` / `wait_agent` / `close_agent` / `resume_agent` (~3,000 LOC). Claude Code has a 7-typed `Task` taxonomy. Koda has `InvokeAgent { background: true | false }` + `agent_name="fork"`. Same coverage, ~10× less surface.

3. **Compile-time tool dispatch via `match`.** The moment a `trait Tool { … }` + `HashMap<String, Box<dyn Tool>>` registry shows up, runtime errors when adding tools become possible. Koda's `match` statement makes that a compile error. Zed and gemini-cli both pay the dynamic-dispatch tax.

---

*Review generated by code-puppy-7be69c against `DESIGN.md` and the source tree at HEAD.*

lijunzh:
## Summary

Closes #1045

Implements the two outstanding gaps from the #1072 sandbox spec. Gap 2 (Auto mode hard-fail) was already closed.

---

### Gap 1 — `allow_git_config` enforcement (default: `false`)

Both backends now block writes to `.git/hooks/` and `.git/config` when the policy flag is off (the default), preventing the two primary git-based sandbox-escape vectors:

- `git config core.fsmonitor <cmd>` / `core.hookspath` — blocked via `(deny file-write* (literal "…/.git/config"))`
- Direct hook creation in `.git/hooks/` — blocked via `(deny file-write* (subpath "…/.git/hooks"))`

| Backend | Mechanism |
|---|---|
| **seatbelt (macOS)** | `git_config_deny_rules(root)` — pure SBPL string builder, follows `credential_deny_rules` pattern. Emitted after credential denies, before policy overlay. Flag is part of the cache key so hits are always correct. |
| **bwrap (Linux)** | `apply_git_config_deny(cmd, root)` — pre-creates `.git/hooks/` so bwrap has a mountpoint, then `--ro-bind`s it. `.git/config` gets `--ro-bind`'d only when it already exists. No-op on non-git directories. |

---

### Gap 2 — Auto mode hard-fail ✅ already closed

---

### Gap 3 — Seed `allow_write` with canonical project root

`policy_for_agent()` removes the `let _ = project_root` stub and sets `fs.allow_write = [canonical_root]` across all trust modes.

Before this fix `allow_write` was always `[]`, making `compose()`'s parent-wins rule vacuously correct but meaningless — `intersect([], []) == []` regardless of what the kernel baseline actually grants. Seeding with the project root makes the lattice match reality.

The `compose_child_policy_with_strict_default_parent_is_just_child_policy` test is updated to document the known non-identity exception: `strict_default()` is the algebraic identity for all fields *except* `allow_write`, where parent-wins semantics make the empty parent list override the child's value. Top-level agents never go through `compose` — they call `policy_for_agent` directly.

---

### Tests added (9 new)

**`koda-sandbox` (seatbelt):**
- `git_config_deny_rules_cover_hooks_and_config`
- `git_config_deny_rules_deny_only_writes_not_reads`
- `build_command_includes_git_deny_rules_when_not_allowed`
- `build_command_omits_git_deny_rules_when_explicitly_allowed`

**`koda-sandbox` (bwrap, Linux-only):**
- `git_config_deny_skips_non_git_dir`
- `git_config_deny_adds_ro_bind_for_existing_git_dir`
- `build_command_skips_git_deny_when_allowed`

**`koda-core`:**
- `policy_for_agent_seeds_allow_write_with_canonical_root`
- `policy_for_agent_allow_write_survives_nonexistent_root`

### DIFF
--- a/koda-core/src/sandbox.rs
+++ b/koda-core/src/sandbox.rs
@@ -212,8 +212,6 @@ pub fn build(
 /// retune. We change values once we have data; we change *where the
 /// values live* now so the change is one-line later.
 pub fn policy_for_agent(trust: crate::trust::TrustMode, project_root: &Path) -> SandboxPolicy {
-    // `project_root` reserved for PR-4+ (seeding fs.allow_write).
-    let _ = project_root;
     let mut policy = SandboxPolicy::strict_default();
     policy.limits.wall_time_secs = Some(match trust {
         crate::trust::TrustMode::Plan
@@ -235,6 +233,26 @@ pub fn policy_for_agent(trust: crate::trust::TrustMode, project_root: &Path) -> 
         crate::trust::TrustMode::Safe => 5,  // user gate exists, balanced
         crate::trust::TrustMode::Auto => 10, // no human gate — max paranoia
     };
+    // Gap 3 of #1072 (PR-4 of #934): seed allow_write with the canonical
+    // project root so the composition lattice accurately reflects what the
+    // kernel baseline already grants.  Without this, compose()'s intersection
+    // rule for allow_write is vacuously correct but has no observable effect:
+    // intersect([], []) == [] regardless of what the baseline sandbox does.
+    //
+    // Canonicalize so symlink-based project roots don't produce a different
+    // key than their realpath counterpart.  Falls back to the raw path on
+    // non-existent roots — sub-agent dispatch can build policy before its
+    // workspace is materialized (the test "does not panic on nonexistent
+    // project root" pins this contract).
+    //
+    // Plan mode is intentionally included: reads dominate in Plan but writes
+    // ARE permitted within the project (e.g. writing scratch files).  The
+    // higher-layer tool registry is what enforces read-only semantics for
+    // Plan, not the kernel sandbox write-allow list.
+    let canonical_root = project_root
+        .canonicalize()
+        .unwrap_or_else(|_| project_root.to_path_buf());
+    policy.fs.allow_write = vec![canonical_root.into()];
     policy
 }
 
@@ -1177,18 +1195,97 @@ mod tests {
 
     #[test]
     fn compose_child_policy_with_strict_default_parent_is_just_child_policy() {
-        // Identity case: when there's no meaningful parent (top-level
-        // invocation, bg-spawned agent), `strict_default()` is the
-        // algebraic identity. The composed policy should equal what
+        // Identity case for *most* fields: when there's no meaningful parent
+        // (top-level invocation, bg-spawned agent), `strict_default()` acts
+        // as the algebraic identity — the composed output matches what
         // `policy_for_agent` would have returned alone.
+        //
+        // Exception — `allow_write`: compose uses parent-wins semantics
+        // ("allows narrow toward parent"), and `strict_default()`'s empty
+        // allow list wins over the child's `[project_root]`.  This is
+        // intentional: a zero-opinion parent shouldn’t grant write permissions
+        // that the parent never explicitly approved.  In practice, top-level
+        // agents never go through `compose_child_policy`; they use
+        // `policy_for_agent` directly.  Sub-agents that DO go through compose
+        // will have a real parent whose `allow_write: [parent_root]` gives
+        // the child's root access via the parent-wins rule, provided the
+        // parent and child share the same project root (the common case).
         let dir = tempfile::tempdir().unwrap();
         let parent = SandboxPolicy::strict_default();
         let composed = compose_child_policy(&parent, crate::trust::TrustMode::Safe, dir.path());
         let child_alone = policy_for_agent(crate::trust::TrustMode::Safe, dir.path());
+        // All fields other than allow_write must match the child's policy.
+        assert_eq!(composed.fs.deny_read, child_alone.fs.deny_read);
         assert_eq!(
-            composed, child_alone,
-            "strict_default parent must be the identity for compose"
+            composed.fs.mandatory_deny_search_depth,
+            child_alone.fs.mandatory_deny_search_depth
         );
+        assert_eq!(
+            composed.fs.allow_git_config,
+            child_alone.fs.allow_git_config
+        );
+        assert_eq!(composed.limits, child_alone.limits);
+        assert_eq!(composed.trust, child_alone.trust);
+        // allow_write: parent-wins on the empty strict_default list — result is
+        // empty regardless of what the child wanted.  Document as the known
+        // non-identity exception, not a bug.
+        assert!(
+            composed.fs.allow_write.is_empty(),
+            "parent-wins: strict_default's empty allow_write beats child's [root]; \
+             top-level agents should NOT go through compose (use policy_for_agent \
+             directly) — #1072 Gap 3 comment"
+        );
+    }
+
+    // ── Gap 3 of #1072: policy_for_agent seeds allow_write ───────────────
+
+    #[test]
+    fn policy_for_agent_seeds_allow_write_with_canonical_root() {
+        // Verifies #1072 Gap 3: the composition lattice only makes sense
+        // if allow_write reflects what the kernel baseline actually grants.
+        // Before this fix, allow_write was always [], making
+        // compose()'s intersection vacuously correct but meaningless.
+        let dir = tempfile::tempdir().unwrap();
+        for mode in [
+            crate::trust::TrustMode::Plan,
+            crate::trust::TrustMode::Safe,
+            crate::trust::TrustMode::Auto,
+        ] {
+            let policy = policy_for_agent(mode, dir.path());
+            assert_eq!(
+                policy.fs.allow_write.len(),
+                1,
+                "policy_for_agent must seed allow_write with the project root \
+                 (one entry) for all trust modes, got {} entries for {mode:?}",
+                policy.fs.allow_write.len()
+            );
+            let want = dir.path().canonicalize().unwrap();
+            assert_eq!(
+                policy.fs.allow_write[0].as_path(),
+                want.as_path(),
+                "seeded allow_write path must be the canonicalized project root"
+            );
+        }
+    }
+
+    #[test]
+    fn policy_for_agent_allow_write_survives_nonexistent_root() {
+        // Regression guard: sub-agent dispatch calls policy_for_agent
+        // before the workspace is materialized on disk.  canonicalize()
+        // fails on missing paths; the fallback to raw path must not panic.
+        let path = std::path::Path::new("/nonexistent/project/root/xyz");
+        for mode in [
+            crate::trust::TrustMode::Plan,
+            crate::trust::TrustMode::Safe,
+            crate::trust::TrustMode::Auto,
+        ] {
+            let policy = policy_for_agent(mode, path);
+            assert_eq!(
+                policy.fs.allow_write.len(),
+                1,
+                "even with a missing root, allow_write must have one entry"
+            );
+        }
     }
 
     #[test]
--- a/koda-sandbox/src/bwrap.rs
+++ b/koda-sandbox/src/bwrap.rs
@@ -82,6 +82,11 @@ pub fn build_command(
 
     let (mut cmd, home) = base_cmd(project_root);
     apply_credential_denies(&mut cmd, &home);
+    // Gap 1 of #1072: protect git hooks + config when allow_git_config=false.
+    if !policy.fs.allow_git_config {
+        let root = project_root.to_string_lossy();
+        apply_git_config_deny(&mut cmd, &root);
+    }
     apply_policy_overlay(&mut cmd, policy)?;
     cmd.args(["--", "sh", "-c", command])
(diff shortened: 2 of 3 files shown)


lijunzh:
## CI note — macOS flake, not related to this PR

The `Test (macos-latest)` failure is **pre-existing** and unrelated to these changes:

```
thread 'bg_agent_iter_counter_advances_via_status_channel' panicked at koda-core/tests/e2e_agent_test.rs:369:
background agent was never registered in the registry within 5s
```

The same test failed on `main` in run [24972496831](https://github.com/lijunzh/koda/actions/runs/24972496831) before this branch was created — it's a timing-sensitive E2E test that races on the macOS GitHub runner. Ubuntu passed cleanly.

Re-triggering CI.

Issue #1074, opened 2026-04-27
Architecture review against DESIGN.md principles (Q2 2026)


lijunzh:
## Architecture review against `DESIGN.md` principles

I did a full audit of the codebase against the principles laid out in `DESIGN.md` (P1: Personal, P2: Simple enough to own alone, P3: Build for the world six months from now). This issue captures the findings so we have a single trackable record of where the code is principle-aligned, where it has drifted, and what's worth resolving.

A companion PR updates `DESIGN.md` to reflect ground truth and document known tensions. This issue is the longer-form analysis the doc edits link back to.

---

### Codebase shape (snapshot)

| Crate | Production LOC | Notes |
|---|---|---|
| `koda-core` | 41,806 | Engine library |
| `koda-cli` | 19,424 | TUI + headless + ACP server |
| `koda-sandbox` | 14,429 | Seatbelt/bwrap kernel sandbox |
| `koda-test-utils` | 748 | Shared test helpers |
| **Total** | **~76,400** | + ~10,100 test LOC across 36 integration test files |

204 `.rs` files total. Test:production ratio ~1:7.5.

---

### What's solidly principle-aligned ✅

| Principle | Evidence |
|---|---|
| P1 "no plugin system" | Zero plugin/DI framework. All 19 built-in tools compiled in via `match` dispatch. |
| P1 "any model, any provider" | 4 `LlmProvider` impls (Anthropic, OpenAI-compat, Gemini, Mock). Provider-neutral interface. |
| P2 "match dispatch over trait registry" | `koda-core/src/tools/mod.rs:609` — pure `match name { ... }` exactly as DESIGN prescribes. |
| P2 "no traits without 2+ impls" | Audited all 8 traits: only `Persistence` is single-impl, and DESIGN explicitly justifies that as a P3 forward-compat carve-out. |
| P3 "loop guard follows Gemini-CLI pattern" | `loop_guard.rs` — consecutive-identical detection + feedback injection, no fingerprinting, no LLM-based stuck detection. |
| P3 "engine = library, no IO" | `koda-core` exposes `EngineSink` trait with 6 impls (Null/Buffering/Test/Cli/Headless/Acp). All output flows through `EngineEvent`. |
| Conversation-first UI | Slash dropdown is help. No `?` overlay, no `/help` modal. Single render path. |
| Sandbox-as-kernel | Always active. Hardened in #1073 (Gaps 1+3). |

#### Trait audit (all 8)

| Trait | # impls | Verdict |
[...]

Issue #1022, opened 2026-04-23
Multi-Agent Execution Review: validate sandbox/tool/agent system for parallel + sequential + bg + fg


lijunzh:
# Multi-Agent Execution Review — Sandbox, Tool, and Agent System Audit

> Comparative review of Koda's multi-agent execution against Codex
> (`codex-rs`), Claude Code, Gemini CLI, Zed, and Code Puppy. Validates
> that the system can support **parallel / sequential / background /
> foreground** execution across **live chat (TUI)**, **task dispatch
> (headless)**, and **multi-agent (master ↔ sub-agents)** scenarios.
>
> Conducted by 4 parallel sub-agent reviews (rust-programmer ×3,
> typescript-reviewer, python-reviewer) on 2026-04-22.

## Executive Summary

The **architecture is sound** and in several places **better than the
reference projects** it set out to learn from. The bugs that exist are
concentrated in **one corner**: the `background: true` path, which was
implemented as a shadow code path rather than as `tokio::spawn` over the
shared `execute_sub_agent`. Fix the four bg-agent issues (trust,
cancellation, drop, runtime) and the system is production-ready for the
three target scenarios.

### What Koda gets right (keep doing)

| Property | Why it matters |
|---|---|
| `KodaAgent` (immutable, `Arc`-shared) ↔ `KodaSession` (per-conv, mutable) | Cleaner than Zed's monolithic `Thread`. Enables genuine parallel sub-agents without `Arc<RwLock<>>` everywhere. |
| **Real `SandboxPolicy::compose()` algebra** with documented per-field rules | Strictly better than Codex's "verbatim runtime override". Chains accumulate restrictions monotonically. |
| **Workspace isolation** via `ClonefileProvider` (macOS APFS, ~3-4× faster) / `GitWorktreeProvider` (Linux) | Neither Codex, Zed, nor Code Puppy has anything equivalent. Parallel write-agents can't trample each other. |
| **Match-statement tool dispatch** | Compile-time exhaustiveness; missing handlers = compile error. Right at the ~20-tool scale. |
| **Engine boundary is a `Serialize` enum** (`EngineEvent`/`EngineCommand`) | Goes *further* than Zed (whose `ThreadEvent` is in-process only). The 430-LOC ACP server is the proof — Zed's equivalent is 3,467 LOC. |
| **Sub-agent result cache** keyed by `(agent_name, prompt_hash)` | Codex re-spawns a full thread every time. Cache hits = $0 + 0 ms. |
[...]

PR #1075, opened 2026-04-27
docs(design): reflect ground truth from architecture audit

lijunzh:
## Summary

Documentation-only update from the architecture audit in #1074. No code changes.

This PR reconciles `DESIGN.md` with what the codebase actually does today, documents previously-undocumented subsystems, and turns implicit principle exceptions into explicit, justified ones.

### Factual drift fixed

| Section | Old claim | Reality |
|---|---|---|
| Engine as a Library | ACP server `~430 LOC` | `server.rs` 501 + `acp_adapter.rs` 554 = `~1,050 LOC` |
| Sub-agent rationale | bg-task surface "in 200 LOC" | `bg_agent.rs` is 831 production LOC (grew in #1022 hardening) |

### New sections added

- **Tool Name Normalization (P2, documented exception to P3)** — explains why `tool_normalize.rs` exists despite P3 saying "fix the model upstream." Cosmetic protocol detail, single boundary point, unambiguous aliases only. Draws the bright line: name canonicalization OK, argument-shape repair / output coercion not OK.

- **Compaction: Two Layers (P2, P3)** — comparison table of microcompact (zero LLM cost, time-triggered) vs full compact (LLM call, threshold-triggered). Acknowledges `COMPACTABLE_TOOLS` dual-case legacy DB defense.

- **Engine-Tracked Progress vs. Model-Tracked Todos (open tension)** — documents the `progress.rs` ↔ `TodoWrite` overlap as a known unresolved tension with a 3-step resolution path (instrument → delete or rename).

### P1 ambiguity closed

DESIGN forbade `--autonomy` but the codebase shipped `--mode` (exact match) plus several model-call parameter flags. Now documented as scenario carve-outs (justified for non-interactive headless use where slash commands aren't reachable), with a "any new flag without a carve-out is a P1 violation" gate at PR review.

## Stats

```
DESIGN.md | 104 +++++++++++++++++++++++++++++++++++++++++++++++++++++++-----
1 file changed, 96 insertions(+), 8 deletions(-)
```

## Why this matters

The whole point of `DESIGN.md` is to be the load-bearing artifact future-you (and external readers via docs.rs) trust to understand the codebase. Drift between doc and code makes the doc actively harmful — readers infer the wrong principles from the wrong examples. This is a periodic honesty pass.

## Closes / refs

- Tracked in #1074 (full audit, including the trait impl audit and file-size review)
- No issue closed by this PR — #1074 stays open as the umbrella for any follow-up code work that comes out of the audit (e.g. resolving the `progress.rs` ↔ `TodoWrite` tension)

## Test plan

`cat DESIGN.md | grep -E '^##|^###'` — section list reads coherently. Paren balance verified by script. Markdown rendering preserved.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -188,6 +188,17 @@ from `KodaSession` (per-conversation, mutable). Zed conflates these into
 shares the agent definition by `Arc` and has its own session for state. No
 `Arc<RwLock<>>` everywhere.
 
+**Known leak: background sub-agent status** (tracked in #1076). The boundary
+claim above — "all engine output flows through `EngineSink`" — is true for
+inference output but not for `BgAgentRegistry` status. The TUI today reads
+bg-task state by calling `BgAgentRegistry::snapshot()` directly through a
+shared `Arc` it grabs out of `KodaSession`. There is no `BgTaskUpdate`
+variant in `EngineEvent`, so ACP clients and headless mode cannot surface
+live bg-task status even though the engine has the data. Fix is to add
+`EngineEvent::BgTaskUpdate { id, status, … }` and have the registry push
+to the sink on transitions (~50–100 LOC). Until then this is a documented
+carve-out, not a permanent design choice.
+
 ### ACP (Agent Client Protocol) (P3)
 
 Koda's server mode will speak ACP. Both Zed and Goose independently converged
@@ -387,29 +398,43 @@ make snake_case unreachable in new sessions. The dual-case list protects
 legacy DB rows from before normalization existed. Acceptable redundancy;
 delete after a full schema migration if/when one happens.
 
-### Engine-Tracked Progress vs. Model-Tracked Todos (open tension)
+### Progress Tracking: Model-Driven Only (P3)
 
-`progress.rs` (engine-side) auto-extracts "files created/edited/deleted" and
-"tests run with exit codes" from tool results, persists them in the DB, and
-re-injects them into the system prompt so the model retains a "done list"
-across compaction.
+Koda tracks "what's been done this session" through a single mechanism: the
+**`TodoWrite` tool** (model-driven, schema copied from Claude Code), backed
+by `FileTracker` for typed file-ownership state used by the auto-approve-delete
+gate.
 
-`TodoWrite` (model-side tool) lets the model write a structured todo list
-to the same DB.
+**Field survey** (none of the reference projects do engine-side progress
+extraction):
 
-**These overlap**. P3 ("let the model drive") argues `TodoWrite` is the
-canonical mechanism and `progress.rs` reimplements verification in the
-engine. The counter-argument: `progress.rs` is *automatic* and survives
-model forgetfulness after compaction, where `TodoWrite` requires the model
-to remember to call it.
+| Project | Engine extracts progress from tool output? | Model-driven todo tool? |
+|---|---|---|
+| Claude Code | No | `TodoWrite` |
+| Codex (OpenAI) | No | `update_plan` |
+| Gemini CLI | No | None (model handles natively) |
+| Zed agent | No | `update_plan`-style |
 
-**Status**: kept for now because models *do* skip `TodoWrite` calls under
-compaction pressure, but flagged for review. Resolution path:
+Koda used to have a third mechanism — `progress.rs`, an engine-side string-
+matcher that scraped `"Created"`/`"Wrote"`/`"Applied"`/`"edited"` patterns
+from tool result text and re-injected a synthesized "done list" into the
+system prompt. It was an exact instance of the P3 anti-pattern
+("don't scaffold around model weakness") and produced two parallel "done"
+lists in the prompt because `TodoWrite` already covered the same ground
+better.
 
-1. Measure how often `TodoWrite` is actually called in long sessions
-2. If frequent enough, delete `progress.rs` (~365 LOC) and rely on the model
-3. If not, keep `progress.rs` but rename it `auto_progress.rs` and document
-   it as a documented engine-side compensator (like `tool_normalize.rs`)
+**The replacement** (tracked in #1077):
+
+1. Improve `compact.rs` summarization prompt to explicitly preserve file
+   paths created/modified and outstanding tasks (~20 LOC).
+2. Add `FileTracker::summary_for_prompt()` and inject it into `prompt.rs`
+   so "files touched this session" comes from the typed dispatch outcome,
+   not from pattern-matched output text (~30 LOC).
+3. Delete `progress.rs` (~365 LOC removed).
+
+Net: ~315 LOC removed, three mechanisms collapse to two (model-owned task
+list + engine-typed file ownership), P3 restored, no string-matching of
+tool output.
 
 ### Sub-Agent Model Routing (P1, P3)
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -66,7 +66,18 @@ and a recompile.
   time, it must be. Flags that select an execution scenario are fine (`-p`
   for headless, `server --stdio` for ACP) — flags that alter behavior within
   a scenario are not (`--autonomy`, `--model-tier`). If something needs to
-  change, change the code
+  change, change the code.
+  
+  *Documented exceptions* (each justified by a non-interactive scenario):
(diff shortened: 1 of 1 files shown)


lijunzh:
Both excellent questions — neither has the obvious answer. Code-grounded analysis below.

---

## Q1: Engine/UI boundary — does sub-agent background management belong in core or CLI?

**Short answer**: lifecycle ✅ already in core, but the **status surface has a real boundary leak** that no one's noticed because the TUI is the only consumer today.

### What's in each crate today

| Concern | File | Crate | Verdict |
|---|---|---|---|
| Lifecycle / state machine (`reserve` → `attach` → `drain_completed` → `cancel`) | `koda-core/src/bg_agent.rs` (831 prod LOC) | core | ✅ Right place |
| Model-callable tools (`ListBackgroundTasks`, `CancelTask`, `WaitTask`) | `koda-core/src/tools/bg_task_tools.rs` | core | ✅ Right place |
| TUI rendering of bg-task status | `koda-cli/src/tui_bg_tasks.rs` (798 LOC, mostly tests) | cli | ✅ Right place |

`koda-core` does **not** depend on `koda-cli` (verified: `grep koda-cli koda-core/Cargo.toml` is empty). Dependency direction is clean.

### The leak: status doesn't flow over `EngineEvent`

The TUI reads bg-task state by calling `BgAgentRegistry::snapshot()` directly through a shared `Arc<BgAgentRegistry>` it gets out of `KodaSession`. **No `BgTaskStarted` / `BgTaskProgress` / `BgTaskCompleted` variant exists in `EngineEvent`.** Verified:

```bash
$ grep "BgTask\|BgAgent" koda-core/src/engine/event.rs
# (empty)
```

Consequences:
1. **ACP clients see nothing** — `acp_adapter.rs` has zero references to `BgAgent`/`BgTask`. An IDE talking to koda over ACP can't surface a live "3 background agents running" panel even though the engine has the data.
2. **Headless mode sees nothing** — same reason. A headless `--prompt` invocation that spawns a background sub-agent can't report status to whoever's tailing stdout.
3. **The TUI works because it shares the process** with the engine and grabs the `Arc` directly, sidestepping the documented boundary.

This is a `DESIGN.md` claim that's *almost* true — "all output flows through `EngineSink`" is accurate for inference output but not for bg-task status. The doc should either:
- **(a) Acknowledge the carve-out**: "bg-task status is read via direct `Arc` from in-process clients only; ACP clients get a polled snapshot via a future `bg_status` request" — and ship that request, or
- **(b) Fix it properly**: add `EngineEvent::BgTaskUpdate { id, status, ... }` and have `BgAgentRegistry` push to the sink when status changes. ~50-100 LOC, restores the invariant.

I'd vote (b). The cost is small and the boundary cleanliness is exactly the thing that makes ACP cheap (`~1,050 LOC` ACP surface vs Zed's `~3,467`). Let it leak and the next ACP feature pays the cost.

### Recommended doc/code work
- **Doc** (this PR or follow-up): note the current bg-status carve-out
- **Code** (separate issue): add `EngineEvent::BgTaskUpdate` + plumb from registry → sink. Probably half a day.

---

## Q2: Two progress trackers — and what do CC / Codex / Gemini CLI / Zed do?

It's actually **three** mechanisms, not two. And yes, this is over-engineered. The fix is clearer than I thought yesterday.

### What koda has today

| # | Mechanism | Who writes | How it captures state | Where it lives |
|---|---|---|---|---|
| 1 | `progress.rs` | Engine (auto) | **String-matches tool result text** for `"Created"`, `"Wrote"`, `"Applied"`, `"edited"`, `"replacement"` | `koda-core/src/progress.rs` (365 LOC) |
| 2 | `TodoWrite` tool | Model (explicit call) | Model rewrites full task list each call; schema = CC's TodoWrite | `koda-core/src/tools/todo.rs` |
| 3 | `FileTracker` | Engine (typed) | Records file ownership when `Write`/`Edit`/`Delete` runs successfully — uses the actual tool dispatch outcome, not pattern-matched output | `koda-core/src/file_tracker.rs` |

Both #1 and #2 inject into the system prompt every turn. The model literally sees two parallel "what's been done" lists. #3 is accurate and already exists but **isn't injected** — it's only consumed by the auto-approve-delete logic.

### What the field does

| Project | Engine-driven progress extraction? | Model-driven todo tool? | Notes |
|---|---|---|---|
| **Claude Code** | ❌ No | ✅ `TodoWrite` (the one koda copied) | Model owns the task list. Engine does compaction + `microCompact`, neither of which auto-extracts "files touched". |
| **Codex (OpenAI)** | ❌ No | ✅ `update_plan` | Same pattern as CC — model-driven plan tool, no engine-side parsing of tool results. |
| **Gemini CLI** | ❌ No | ❌ No (relies on context window + the model) | Gemini CLI is the most minimal — no plan tool, no engine progress. Model handles it natively. |
| **Zed agent** | ❌ No | ✅ `update_plan`-style tool on the thread | `Thread` entity manages messages; progress is a model concern. |
| **Koda** | ✅ `progress.rs` | ✅ `TodoWrite` | **Only project of the four with engine-side auto-extraction.** P3 violation. |

**Koda is the outlier.** All four reference projects landed on "model owns the task list, engine doesn't second-guess." `progress.rs` is a pattern none of them adopted.

### Why does koda have it then?

The defensible reason: models occasionally skip `TodoWrite` calls when the context gets tight, and `compact.rs` summarization can lose specific file paths. So `progress.rs` is "engine compensation for unreliable model behavior." Which is **literally the P3 anti-pattern DESIGN warns against**:

> "Frontier models won't need this in 6 months. Don't scaffold around weakness."

### The right fix

It's actually a three-step simplification, not a one-step delete:

1. **Improve the compaction prompt** — explicitly instruct the summarizer to preserve the list of file paths created/modified and outstanding tasks. CC does this; we should too. (Maybe 20 LOC change in `compact.rs` prompt template.)

2. **Promote `FileTracker` to a prompt source** — it already has accurate, typed file ownership via `tool_dispatch.rs` outcomes. Adding `FileTracker::summary_for_prompt()` and wiring it into `prompt.rs` gives us "files touched this session" without string-matching tool output. ~30 LOC.

3. **Delete `progress.rs`** — once #1 and #2 land, it has no consumers worth keeping. -365 LOC.

Net: **~50 LOC added, 365 LOC deleted, two prompt sections collapse to one, P3 restored.**

### Why not just delete `progress.rs` today?

Because if you delete it cold without #1 and #2, you might regress the "model loses file context after compaction" case. The instrument-first approach in the current DESIGN.md tension section (count `TodoWrite` calls in long sessions) is fine but actually slower than just doing the right thing — `FileTracker` is already accurate and `TodoWrite` already handles tasks. The data is right there.

---

## Proposed disposition for this PR

This PR documents the current state honestly. I think we should:

- **Merge as-is** — it tells the truth about today's code, which is the main value
- **Open follow-up issues** for the two real problems this analysis surfaced:
  - **Issue A**: `EngineEvent::BgTaskUpdate` — close the boundary leak so ACP/headless can surface bg-task status
  - **Issue B**: Collapse three progress mechanisms into one (`FileTracker` injection + improved compact prompt + delete `progress.rs`)

Both are bigger than this doc PR's scope but small enough to be one-PR each. Want me to open them?

lijunzh:
Updated DESIGN.md with the agreed direction. New commit: 6178b49.

**Engine/UI boundary** \u2014 'Engine as a Library' section now documents the bg-task status leak as a temporary carve-out tracked in #1076. The fix variant (`EngineEvent::BgTaskUpdate`) and the `BgAgentRegistry` sink wiring are spec'd there.

**Progress tracking** \u2014 'Engine-Tracked Progress vs. Model-Tracked Todos (open tension)' section retitled and rewritten as 'Progress Tracking: Model-Driven Only (P3)' with the field survey table (CC/Codex/Gemini-CLI/Zed all model-driven, koda the outlier) and the 3-step fix path tracked in #1077 (-315 LOC net).

Follow-up issues opened:
- #1076 \u2014 Add `EngineEvent::BgTaskUpdate` to close engine/UI boundary leak (`design-debt`, `principles`)
- #1077 \u2014 Collapse three progress-tracking mechanisms into one (`design-debt`, `principles`)

This PR is now ready to merge as-is. The umbrella audit issue #1074, both follow-up issues, and this PR all carry the `design-debt` + `principles` labels for easy filtering later.

Issue #1094, opened 2026-04-27
List tool output omits directory header — parallel calls become unattributable


lijunzh:
## Summary

The `List` tool's output omits the path of the directory being listed and, in non-recursive mode, only emits entry **basenames** with no parent context. When the model fires several `List` calls in parallel, it gets back a stack of unattributed listings and can't reliably tell which output corresponds to which input.

## Reproduction

Session: `6a71c2f3-bf94-4c0e-8709-3056f577b7bd` (model: `minimax-m2.7`, project: `/Users/lijun`, 2026-04-27).

Three parallel `List` calls produced these outputs, in order:

```
Output A:
(empty directory)

Output B:
d monitor
d math_puzzles
  .DS_Store
d dotfiles
d openclaw
d claude_code_src
... (etc.)

Output C:
2 files matched:
.DS_Store
.localized
```

Nothing in any of these outputs identifies which directory was listed. The model has to guess from request order — fine in single-call use, brittle when called in parallel and outright misleading when one call returns `(empty directory)` (which directory was empty? `Documents`? `Downloads`? something else?).

A second instance from the same session listing `/Users/lijun`:

```
.claude.json.backup
d .config
d Music
d .code_puppy
d .gemini
...
  tax
  .zsh_history
d Desktop
...
```

Again — no header identifying the directory, and entries are unsorted basenames only.

## Root cause

In [`koda-core/src/tools/file_tools.rs:617-627`](koda-core/src/tools/file_tools.rs:617) (non-recursive branch):

```rust
let mut reader = tokio::fs::read_dir(&resolved).await?;
while let Some(entry) = reader.next_entry().await? {
    let ft = entry.file_type().await?;
    let prefix = if ft.is_dir() { "d " } else { "  " };
    entries.push(format!("{prefix}{}", entry.file_name().to_string_lossy()));
    ...
}
```

Entries are pushed as bare basenames. The output assembled at line 629–638 is just `entries.join("\n")` (or `"(empty directory)"`) — no header, no `resolved` path included anywhere.

The recursive branch at line 600–615 emits paths relative to `project_root` rather than to `resolved`, so a recursive listing of a subdirectory still doesn't visibly tell the model which directory was listed.

## Suggested fix

Prefix every `List` output with a one-line header naming the listed directory, e.g.:

```
Listing: /Users/lijun/repo (17 entries)
d monitor
d math_puzzles
...
```

And for the empty case:

```
Listing: /Users/lijun/Documents
(empty directory)
[...]

PR #1099, opened 2026-04-28
fix(tui): tool headers show actual path, not always '.'

lijunzh:
## Problem

When a sub-agent (or the main agent) called `List`, `Glob`, or `Grep` with a non-default directory, the tool header rendered the *default* path instead of the actual one. From a user repro of the explore sub-agent walking a multi-package workspace:

```
● InvokeAgent explore  🤖 Sub-agent: explore
● List .
● List .
● Glob */.git
● Glob */Cargo.toml
● Glob */package.json
● List .
● List .
● List .
● List .
● List .
● List .
● List .
...
```

The model was actually calling `List { file_path: "koda-core/src" }`, `List { file_path: "koda-cli/src" }`, `List { file_path: "tests" }`, … — distinct directories. The TUI rendered every single one as `● List .`, making the trace look like the agent was loop-spinning on the project root. Same for `Glob '*/.git' in koda-core/` rendering identically to `Glob '*/.git'` from project root.

## Root cause

`tool_header.rs` looked for `directory` / `path` keys when building the rendered detail:

```rust
let dir = first_string(args, &["directory", "path"]).unwrap_or_else(|| ".".to_string());
```

But every koda built-in tool's schema uses `file_path` for the directory parameter (and has since day one — `tools/file_tools.rs`, `tools/glob_tool.rs`, `tools/grep.rs`). The renderer's keys were obsolete from the moment the module was added (#952, 8 days ago). The lookup never matched, so it always fell back to `.`.

Affected tools:

| Tool   | Schema key  | Renderer looked for     | Result                                  |
|--------|-------------|-------------------------|-----------------------------------------|
| `List` | `file_path` | `directory`, `path`     | always rendered `.`                     |
| `Grep` | `file_path` | `directory`, `path`     | always rendered `... in .`              |
| `Glob` | `file_path` | *(not checked at all)*  | base directory always silently dropped  |

## Fix

Update `list_detail`, `grep_detail`, `glob_detail` (and their `_text` mirrors used by transcript export) to look at the schema's actual key — `file_path` — first, with `path` and `directory` kept as fallbacks for paranoia / future tools.

Bonus UX win: `glob_detail` now surfaces `file_path` when set (`Glob '*.toml' in koda-core/`) and stays terse when omitted (`Glob '*.toml'`) so the common project-root case doesn't gain noise.

## Tests

Four new regression tests, plus a defensive contract test that pins the renderer's key list to the dispatch's key list across all 7 path-bearing tools (`List`, `Grep`, `Glob`, `Read`, `Write`, `Edit`, `Delete`). If anyone renames `file_path` in a tool schema or adds a new path-bearing tool without updating `tool_header.rs`, the build fails instead of the live header silently lying.

- `grep_uses_file_path_key_from_schema` — direct repro
- `list_uses_file_path_key_from_schema` — direct repro (the user's `● List .` storm)
- `glob_surfaces_file_path_when_present` — Glob's optional base now renders
- `glob_omits_base_when_default_root` — keep the common case tight
- `path_bearing_tools_render_actual_dispatch_key` — cross-tool drift catcher

## Verification

- ✅ `cargo test --workspace --features test-support` — 2196 passed (2191 prior + 5 new), 0 failed
- ✅ `cargo clippy --workspace --features test-support --all-targets -- -D warnings` — clean

🐶 generated by code puppy

### DIFF
--- a/koda-cli/src/tool_header.rs
+++ b/koda-cli/src/tool_header.rs
@@ -656,4 +656,114 @@ mod tests {
         let out = detail_text("MysteryTool", &json!({"count": 42}), 80);
         assert_eq!(out, "");
     }
+
+    // ── cross-layer: dispatch body ↔ TUI header agreement ────────────────
+    //
+    // koda has multiple independent formatters for tool calls:
+    //
+    //   1. The dispatch's response body (what the LLM sees)
+    //   2. The TUI header (what the human sees live)
+    //   3. The transcript export (what the human saves to clipboard)
+    //   4. The history replay (what the human sees on session resume)
+    //
+    // Each was written independently and **none share code**. When the
+    // `List` tool's directory parameter standardized on `file_path`,
+    // only the dispatch was updated. The body formatter got fixed in
+    // PR #1096; the TUI / transcript layers got fixed in PR #1099 —
+    // same root symptom ("can't tell which directory was listed")
+    // reported twice from two different angles, because no test had
+    // pinned the two layers together.
+    //
+    // This test does. For a real `List` call against a tempdir, the
+    // body that ships to the LLM and the header that ships to the
+    // user must both surface the same path. If anyone changes the
+    // dispatch's arg key without touching `tool_header.rs`, this
+    // test fails before the user ever sees a misleading `● List .`.
+    //
+    // Glob and Grep aren't end-to-end-tested here (they need a
+    // `koda_sandbox::fs::FileSystem` impl that koda-cli doesn't depend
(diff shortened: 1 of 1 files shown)


lijunzh:
Per discussion, added option (b) — two cross-layer tests in `tool_header.rs` (commit 75aea53) that pin the LLM-facing body and the user-facing header together. They catch the bug class behind both #1094 and this PR.

Option (c) — the proper structural fix (single `ToolCallSummary` formatter shared by all rendering layers) — is tracked in #1100.

Issue #1109, opened 2026-04-28
test: systemic data-race patterns causing recurring CI flakes (env vars, current_thread runtime, polling sleeps)


lijunzh:
## Summary

Recent CI flakes (#1090 most recently, #211 earlier) are not isolated incidents — they're symptoms of a few systemic patterns in our test suite that keep producing data races and timing-dependent failures. This issue tracks the patterns and proposes principled fixes so we can stop playing whack-a-mole.

### F1 — Process-global env var mutation as test fixture (highest leverage)

`KODA_MOCK_RESPONSES` (and friends like `XDG_CONFIG_HOME`, `HTTP_PROXY`, `STAGE2_BIN_ENV_KEY`, `HYPERLINK_KILL_SWITCH`) are mutated from tests via `unsafe { std::env::set_var() }` / `remove_var()`. In Rust 2024 edition this is **undefined behavior** in any multi-threaded program, which `#[tokio::test]` is.

The current mitigation is `koda_test_utils::ENV_MUTEX` — a single tokio mutex that every test is *expected by convention* to acquire before touching env. A new test that forgets the lock silently reintroduces UB.

Hits include:

- `koda-core/tests/e2e_agent_test.rs` — 8 separate `set_var` / `remove_var` pairs around `KODA_MOCK_RESPONSES`.
- `koda-core/tests/http_client_config_test.rs` — proxy env vars.
- `koda-core/src/db/tests.rs:1133` — `XDG_CONFIG_HOME`.
- `koda-core/src/providers/mock.rs:333` — a **second, separate** `ENV_MUTEX` (std::sync::Mutex) inside the lib-tests module. Different binary so it doesn't collide today, but it's a "two obvious ways to do it" anti-pattern (Zen of Python violation) and a footgun the next time these tests get refactored together.

**Why this keeps biting us**: #211 was filed and "fixed" by adding the mutex; #1102 just fixed an unrelated bg-agent issue but the env-var pattern is still everywhere.

**Proposed fix**:
1. We already have `koda_core::runtime_env` — a thread-safe `RwLock<HashMap>` built explicitly to replace `unsafe { set_var }`. Migrate `KODA_MOCK_RESPONSES` (and any other test-only knob) to read from `runtime_env::get` instead of `std::env::var`. Then tests can `runtime_env::set` without any unsafe / mutex / convention.
[...]

PR #1112, opened 2026-04-28
feat(transcript): persist + fold sub-agent traces and engine events (#1108 P1b/P2a)

### LINKED ISSUE #1108: Transcript export drops tool names + args (schema mismatch) and many sink-only events, hurting debuggability
## TL;DR

While debugging a parallel sub-agent session via the markdown transcript export (`/export` / `transcript.rs`), I discovered that **every tool call in every exported transcript is rendered as `### 🔧 **Tool**` with no name and no arguments**. Existing tests pass because their fixtures use a JSON shape that doesn't match what gets persisted in production.

Beyond that schema bug, several categories of debugging-grade information are emitted to the sink only and never reach the DB / transcript at all.

### Symptom

In a real session export, parallel `InvokeAgent` calls render as:

```
### 🔧 **Tool**
### 🔧 **Tool**
### 🔧 **Tool**
```

No tool name, no args. The reader has to *infer* `InvokeAgent` vs `WaitTask` from the shape of the output JSON. For a parallel call, you also can't tell which output belongs to which invocation.

### Root cause

Schema mismatch between persistence and rendering:

- `ToolCall` ([koda-core/src/providers/mod.rs:57](koda-core/src/providers/mod.rs)) is a flat struct:
  ```rust
  pub struct ToolCall {
      pub id: String,
      pub function_name: String,   // ← flat
      pub arguments: String,
      pub thought_signature: Option<String>,
  }
  ```
  `serde_json::to_string` produces `{\"id\":..., \"function_name\":\"InvokeAgent\", \"arguments\":\"{...}\"}`. This is what `inference.rs:974` writes to the `messages.tool_calls` column.

- `transcript.rs:181-182` reads the **nested OpenAI shape** instead:
  ```rust
  let name = call[\"function\"][\"name\"].as_str().unwrap_or(\"Tool\");
  let args_json = call[\"function\"][\"arguments\"].as_str().unwrap_or(\"{}\");
  ```
  Both lookups miss and fall through to the defaults — hence `**Tool**` and empty args on every export.

### Why tests didn't catch it

The transcript tests at `transcript.rs:488` and `:510` build fixtures using the nested shape (`{\"function\": {\"name\": \"Read\", \"arguments\": ...}}`) instead of the flat shape that `serde_json::to_string(&tool_calls)` actually produces. So the unit tests pass while every real export is broken.

### Fix

1. Change `transcript.rs:181-182` to read the flat shape (`call[\"function_name\"]`, `call[\"arguments\"]`).
2. Update the test fixtures at `:488`, `:510`, and `:632` to round-trip through `serde_json::to_string(&Vec<ToolCall>)` so they match production data.
3. Add a regression test that exports a transcript built from real `ToolCall` structs and asserts the tool name appears in the output.

## P1 — Sink-only events never reach the DB

`EngineEvent::Info` and `EngineEvent::BgTaskUpdate` are emitted to the live sink (TUI / ACP) but never written to the database — see emissions at `inference.rs:645, 698, 1253, 1259` and the bg-agent status fan-out at `bg_agent.rs:295-301`. The transcript export reads only `messages` rows, so it misses:

- `✅ Background agent 'X' completed/failed` lines, including the `BufferingSink` narrative trace that captures the sub-agent's tool calls
- Bg-agent status transitions (`Pending → Running { iter: 1 } → ... → terminal`)
- Microcompact summaries (`🧹 cleared N old tool results`)
- Loop-detector nudges
- Rate-limit retry warnings (`waiting 4s, retrying...`)

For background sub-agents specifically, this is the most important debugging signal — *what was each task doing during the wait window* — and it's completely absent from the export.

### Suggested fix

Add a `session_events` table (id, session_id, ts, kind, payload) and persist these event types alongside messages. The transcript renderer can then interleave them by timestamp.

## P1 — `tool_call_id` not surfaced

`Message.tool_call_id` exists on each tool result row and is used internally for correlation, but the markdown export never prints it. Under parallel tool calls (e.g. 3× `InvokeAgent` in one assistant turn), three `Tool` headers are followed by three `Output:` blocks with no way to match them up. Surfacing `[call_abc123]` next to the tool header would make parallel transcripts navigable.

## P2 — Sub-agent inner trace not persisted

When `InvokeAgent` runs (foreground or background), the sub-agent has its own `inference_loop` — but its inner history lives in a transient `BufferingSink` and only surfaces as a single `Info` event on completion. Nothing about the sub-agent's run lands in the DB. For the failed bg agent in my session, this is why `Result: Error: Failed to call LLM API` was the *entire* signal: no HTTP status, no iteration count at failure time, no record of the sub-agent's tool calls before it died.

### Suggested fix

Persist `BgAgentResult.events` (and the equivalent for foreground sub-agents) in a child column or a related table, then render as a collapsible block under the parent's tool result.

## P2 — Upstream API errors stripped

The user-injected failure message reads `Result: Error: Failed to call LLM API` — no HTTP status code, no provider response body, no error class. For triaging \"is this rate limit, auth, or 5xx,\" the upstream error class needs to flow through.

## P3 — Truncated output not recoverable from export

`SUMMARY_RESULT_LINES` caps tool result output and appends `… (N more lines)`. The unbounded text is in `messages.full_content` but never exposed by the export — even `--verbose` only raises the cap. An `--unbounded` mode that pulls `full_content` would let users debug long outputs without re-running.

## Priority summary

| Priority | Fix | Effort |
|---|---|---|
| P0 | Fix `function_name` / `function.name` schema mismatch + update test fixtures to round-trip real `ToolCall` JSON | trivial |
| P1 | Persist `Info` / `BgTaskUpdate` events to a `session_events` table; render inline | medium |
| P1 | Surface `tool_call_id` in the markdown tool header | trivial |
| P2 | Persist sub-agent inner trace (`BgAgentResult.events`) and render collapsibly | medium |
| P2 | Carry upstream API error class/body into bg-agent failure injection | small |
| P3 | Add `--unbounded` export mode that pulls `full_content` | small |

## Why this matters

I tried to evaluate whether koda's parallel sub-agent system was implemented correctly using a real session transcript ([koda#1106](https://github.com/lijunzh/koda/issues/1106) for the findings). The transcript made that significantly harder than it needed to be: the most useful evidence (which agent did what, with what args, and what the failed agent saw before dying) was either silently dropped (P0) or never persisted (P1+). Fixing P0 alone would unblock most retrospective debugging from saved transcripts.

lijunzh:
## Summary

Closes the second half of #1108. PR #1111 (P1a) made tool calls visible in the markdown export by surfacing tool name, args, and call_id. This PR addresses the deeper gap: bg sub-agents and top-level engine events were sink-only — neither the live TUI trace nor the exported transcript could tell you what a bg agent did or when context was compacted.

### P1b — engine event persistence

- New `session_events` table: `(id, session_id, kind, payload, parent_tool_call_id, created_at)`. Append-only, indexed by session_id. Three event kinds: `info`, `bg_task_update`, `sub_agent_event`.
- `Persistence::insert_session_event` / `load_session_events` trait methods. Fire-and-forget at the call site — a DB hiccup must not crash a turn just to persist a trace line.
- `PersistingSink` decorator wraps any `EngineSink` and forks Info / BgTaskUpdate events to both the underlying sink AND the DB. Top-level inference uses it; foreground sub-agents don't (their events flow inline through the parent).

### P2a — sub-agent trace folding

- `BgAgentResult` carries the parent's `InvokeAgent` tool_call_id, threaded from `tool_dispatch::execute_one_tool` → `execute_sub_agent` → `BgAgentRegistry::attach` → `drain_completed`.
- Inference loop's drain handler persists each captured bg-agent trace line as a `sub_agent_event` row keyed by that call_id.
- Transcript renderer buckets events by `parent_tool_call_id`: parented events fold into a `<details>` block under the matching Tool result; top-level events surface in a new 'Background activity' section at the end of the transcript.

## Why two render targets, not one

Sub-agent events are correlated with a specific tool call — folding them under that call's result makes the transcript scan-able. Top-level events (microcompact, rate-limit retries, bg-task state transitions) have no such anchor; chronological tail section is the honest place for them.

## Tests

- 3 new DB tests: round-trip ordering, per-session isolation, empty-session no-op.
- 5 new transcript renderer tests: fold-under-tool-result, orphan drop, background-activity section, no-events skip, malformed-JSON fallback.
- **1,932 tests pass total. Zero regressions. Clippy clean. `cargo fmt` clean.**

## Files changed

| File | Purpose |
|---|---|
| `koda-core/src/db/{mod,queries,tests}.rs` | schema + trait methods + tests |
| `koda-core/src/persistence.rs` | `SessionEvent` type + `session_event_kind` constants |
| `koda-core/src/engine/sink.rs` | `PersistingSink` decorator |
| `koda-core/src/inference.rs` | drain handler persists trace |
| `koda-core/src/bg_agent.rs` | `parent_tool_call_id` on entry |
| `koda-core/src/sub_agent_dispatch.rs` | wires it through |
| `koda-core/src/tool_dispatch.rs` | passes `Some(&tc.id)` |
| `koda-core/src/tools/bg_task_tools.rs` | pattern update for new field |
| `koda-cli/src/transcript.rs` | bucketing + render |
| `koda-cli/src/tui_commands.rs` | load events for export |
| `koda-cli/src/tui_bg_tasks.rs` | test signature update |

Closes #1108.

🐶 _Authored by code-puppy._

### DIFF
--- a/koda-cli/src/transcript.rs
+++ b/koda-cli/src/transcript.rs
@@ -34,7 +34,7 @@
 //! (full tool output shown)
 //! ```
 
-use koda_core::persistence::{Message, Role};
+use koda_core::persistence::{Message, Role, SessionEvent, session_event_kind};
 use koda_core::tools::{ToolEffect, classify_tool};
 use std::collections::HashMap;
 
@@ -137,9 +137,26 @@ fn extract_tool_call_meta(call: &serde_json::Value) -> (String, String, String) 
 ///   all tool output, session metadata header.
 /// - `verbose = false` (`--summary`): concise, human-readable.
 ///
+/// `events` carries non-message engine events persisted to the
+/// `session_events` table (#1108 P1b/P2a). They split into two
+/// rendering buckets:
+/// - **Sub-agent events** (`parent_tool_call_id = Some(id)`) are
+///   folded as a collapsible `<details>` block under the matching
+///   `Tool` result. Pass an empty slice for sessions with no bg
+///   sub-agents.
+/// - **Top-level events** (`parent_tool_call_id = None`) are appended
+///   in a chronological "Background activity" section at the end,
+///   so the reader can correlate microcompact / rate-limit / bg-task
+///   transitions with the conversation above.
+///
 /// Returns the transcript as a `String`. The caller is responsible for
 /// writing it to the clipboard or a file.
-pub fn render(messages: &[Message], meta: &SessionMeta, verbose: bool) -> String {
+pub fn render(
+    messages: &[Message],
+    events: &[SessionEvent],
+    meta: &SessionMeta,
+    verbose: bool,
+) -> String {
     let mut out = String::with_capacity(if verbose { 16384 } else { 4096 });
 
     // Header
@@ -164,6 +181,19 @@ pub fn render(messages: &[Message], meta: &SessionMeta, verbose: bool) -> String
                     tool_id_to_name.insert(id, name);
                 }
             }
+        }
+    }
+
+    // #1108 P2a: bucket sub-agent events by parent tool_call_id so
+    // each Tool result can render its folded trace in O(1). Top-level
+    // events (no parent) accumulate in a separate Vec for the
+    // "Background activity" tail section. Single pass, two buckets.
+    let mut events_by_parent: HashMap<&str, Vec<&SessionEvent>> = HashMap::new();
+    let mut top_level_events: Vec<&SessionEvent> = Vec::new();
+    for ev in events {
+        match ev.parent_tool_call_id.as_deref() {
+            Some(parent) => events_by_parent.entry(parent).or_default().push(ev),
+            None => top_level_events.push(ev),
         }
     }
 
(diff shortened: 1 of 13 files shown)


lijunzh:
Converted to draft while we tackle #1109 first — the failing test (`bg_agent_iter_counter_advances_via_status_channel`) is a known instance of #1109 F3 (polling sleeps as synchronization). Rather than bandaid this one test, fixing #1109 systematically eliminates the whole category. Will rebase + ready this PR once #1109 lands.

Issue #1141, opened 2026-04-29
Epic: TUI responsiveness & terminal hygiene — v0.3.x graduation blocker


lijunzh:
# 🚀 Epic: TUI responsiveness & terminal hygiene — v0.3.x graduation blocker

## Why this epic exists

Through a comparative review of [`codex`](https://github.com/openai/codex) (also Rust, also ratatui-based) we identified **a single class of bug** that is the largest blocker for koda graduating into v0.3.x:

> Under heavy event throughput (sub-agent fan-out, fast streaming, parallel tools) the TUI's tokio executor is monopolised by engine event processing, causing:
> 1. Composer typing freezes for seconds at a time
> 2. Mouse scroll-wheel events leak as raw escape sequences (`^[[<64;3;7M`) into the prompt — a regression of #540
> 3. Ctrl+C cancellation lags by seconds
> 4. Wasted CPU on redundant redraws

Codex does not have any of these symptoms. The reasons are **architectural**, not patches. This epic ports the relevant ideas back into koda, ranked by ROI.

## Architectural diff (codex vs koda)

| | **codex** | **koda** | Issue |
|---|---|---|---|
| Mouse capture | ❌ Never enabled, uses `?1007h` alternate scroll | ✅ Enabled, custom scroll/select | #1140 |
| Frame scheduling | Coalescing scheduler @ max 120 FPS | Sync `terminal.draw()` per loop iter | #1138 |
| Event-loop fairness | Round-robin polling each iteration | `tokio::select! { biased; ... }` only | #1139 |
| Engine event drain | Naturally bounded (per-poll) | **Unbounded** `while let Ok(...)` | #1137 |
| Process model | TUI ↔ `app-server` over IPC | All in one process | (out of scope — see "Deferred") |
| Composer widget | Hand-rolled 2660-line TextArea | `ratatui-textarea` 0.9 dep | #1116 |
| Alt screen | Inline default, opt-in overlays | Always alt-screen | (deferred follow-up) |

### 🔥 Phase 1 — Stop the bleeding (target: this week)

- [ ] **#1137** — TUI: input freeze + Ctrl+C lag + mouse escape leakage during heavy inference (regression of #540)
  - Bound the unbounded `while let Ok(extra) = ui_rx.try_recv()` drain loop in `tui_handlers_inference.rs`
  - **Cost: XS** — small change, high-impact
  - **Owner:** TBD

[...]

Issue #1140, opened 2026-04-29
TUI: drop EnableMouseCapture, use ?1007h alternate scroll mode (codex parity)


lijunzh:
## Context

The mouse-event-leakage class of bugs (#540 original, regressed in #1137) exists because koda calls `EnableMouseCapture` at startup. Mouse capture turns scroll wheel + clicks into byte sequences crossterm parses as `MouseEvent`s — but if the event loop ever falls behind, partial sequences leak as raw escape codes into the composer.

`../codex` made the opposite trade-off: they **never enable mouse capture**, period. For scrolling they enable ANSI **alternate scroll mode** (`\x1b[?1007h`) instead. The terminal emulator handles wheel scrolling natively (translating wheel → ↑/↓ keys), so:

1. The app never receives mouse events → no leak class possible by construction
2. OS-native text selection works in the terminal
3. No coordinate math, no scroll buffer scroll handling

```rust
// codex-rs/tui/src/tui.rs:271-294 — note the absence of EnableMouseCapture
pub fn set_modes() -> Result<()> {
    execute!(stdout(), EnableBracketedPaste)?;
    enable_raw_mode()?;
    if !keyboard_enhancement_disabled() {
        let _ = execute!(stdout(), PushKeyboardEnhancementFlags(...));
    }
    let _ = execute!(stdout(), EnableFocusChange);
    Ok(())
}

// codex-rs/tui/src/tui.rs:298-315
struct EnableAlternateScroll;
impl Command for EnableAlternateScroll {
    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
        write!(f, "\x1b[?1007h")
    }
}
```

Codex also drops mouse events at the source as a belt-and-braces measure (`event_stream.rs:240-256` returns `None` for unmapped events including mouse).

## What koda loses

We currently use mouse capture for two features in `koda-cli/src/tui_context/events.rs:202-296` and `mouse_select.rs`:

1. **Scroll wheel scrolls the history panel** — replaced by terminal-native scrollback when we leave alt-screen, or by `?1007h` while in alt-screen
2. **Click-and-drag text selection w/ clipboard copy** — terminal-native selection works in most terminals (iTerm2, kitty, Ghostty, Alacritty, Windows Terminal, WezTerm, GNOME Terminal). Variable in tmux/screen but `Shift+drag` workaround is well-known and documented.

## What koda gains

- Eliminates the entire mouse-escape-leakage bug class by construction
[...]

Issue #1138, opened 2026-04-29
TUI: extract frame coalescing scheduler — decouple draws from event loop (codex parity)


lijunzh:
## Context

Koda calls `terminal.draw()` synchronously inside every iteration of the inference event loop (`koda-cli/src/tui_handlers_inference.rs:75-94`). Under high event throughput (sub-agent fan-out, fast streaming, parallel tools) that's a redraw per engine event — wasteful CPU + a contributor to the executor monopolisation that causes #1137.

`../codex` decoupled this: redraws are *requested*, not performed inline. A dedicated `FrameRequester` / `FrameScheduler` task coalesces requests at max 120 FPS into a broadcast channel. The main event loop just consumes `TuiEvent::Draw` from the channel like any other event.

```rust
// codex-rs/tui/src/tui/frame_requester.rs:50-95 — coalescing scheduler
async fn run(mut self) {
    let mut next_deadline: Option<Instant> = None;
    loop {
        let target = next_deadline.unwrap_or_else(|| Instant::now() + ONE_YEAR);
        let deadline = tokio::time::sleep_until(target.into());
        tokio::pin!(deadline);
        tokio::select! {
            draw_at = self.receiver.recv() => {
                let draw_at = self.rate_limiter.clamp_deadline(draw_at);
                next_deadline = Some(next_deadline.map_or(draw_at, |cur| cur.min(draw_at)));
                continue;  // fold multiple requests into one frame
            }
            _ = &mut deadline => {
                if next_deadline.is_some() {
                    next_deadline = None;
                    self.rate_limiter.mark_emitted(target);
                    let _ = self.draw_tx.send(());
                }
            }
        }
    }
}
```

## Goal

Replace synchronous `terminal.draw()` calls in koda's hot path with `frame_requester.schedule_frame()`. The actual draw happens on the next `TuiEvent::Draw` that the unified event stream emits.

## Why this matters for v0.3.x

- 1000 `TextDelta`s during a streaming response → today: 1000 redraws → tomorrow: 8 redraws (at 120 FPS)
- Frees the executor between events, hardening #1137 even further
- Sets up the architectural foundation for the codex-style unified event stream
- Preconditions a longer-term inline-rendering experiment (drop alt-screen for the main UI)

## Implementation sketch

[...]

Issue #1116, opened 2026-04-29
Epic: hybrid port + build of composer surface (codex-derived textarea + koda-shaped integration)


lijunzh:
## Status (2026-04-29)

**Activated.** Originally filed as a deferred strategic question with criteria for when to revive; per the strategic discussion (2026-04-29 with @lijunzh), we're now committing to a **3-PR hybrid port + build** approach in service of **#1175** (capability milestone). Background and trigger criteria preserved below for archaeological reasons; **the new plan is at the bottom of this issue.**

---

## Original context (preserved)

During the #1104 P3 polish triage, an auditor recommended bumping `ratatui-textarea` 0.9 → newer. Investigation (PR #1115) revealed:

1. We're already on the latest published `ratatui-textarea` (0.9.1). There is no upgrade target.
2. The auditor likely confused our crate with `tui-textarea` (0.7.0, separate lineage) or `tui-textarea-2` (0.10.2, community fork). Three different crates, three different upstreams.
3. Our composer is the **highest-touch UX surface** in the product — every user interaction with koda flows through it.

## What `../codex` does (preserved)

Codex (`codex-rs/tui/src/bottom_pane/textarea.rs`) hand-rolled a **2,660-line `TextArea`** built directly on `ratatui::widgets::WidgetRef` + `crossterm` key events. Zero `ratatui-textarea` / `tui-textarea` dep. Their reasons (visible in the file's docs and observable from the structure):

| Codex design choice | Why it matters for koda |
|---|---|
| Tight integration with a `TextElement` system (atomic placeholders for slash-commands that move/delete as one unit) | Our wrap_input slash-command and file-mention flows fight the third-party widget today |
| Custom kill-buffer semantics (single-entry, deliberately not Emacs-style) | We've already had to work around `ratatui-textarea`'s assumptions in several places |
| No third-party TUI-input dep risk | Exactly the deferral reasoning that drove #1104's item #4 — they solved it structurally |
| Composer = product | For an interactive coding agent, input handling is core, not a vendored widget |

## What changed: the strategic discussion

Original framing was "stay on `ratatui-textarea`, defer until criteria triggered." On 2026-04-29 we re-evaluated and noted:

1. **All four codex-only features are koda-relevant** — image paste, slash autocomplete, history scrollback, paste-burst handling are direct dogfood UX wins.
[...]

PR #1142, opened 2026-04-29
tui(inference): bound drain + round-robin select fairness (#1137, #1139)

### LINKED ISSUE #1137: TUI: input freeze + Ctrl+C lag + mouse escape leakage during heavy inference (regression of #540)
## Symptom

During inference (especially sub-agent fan-out, parallel tools, or fast streaming), the TUI becomes unresponsive:

- Typing in the composer is delayed by seconds
- Mouse scroll-wheel events leak as raw escape sequences (`^[[<64;3;7M`) into the composer — a regression of #540
- Ctrl+C takes a long time to cancel the in-flight turn

Reported live by user observing recent sessions; partially captured in session log `koda-95112.log` (Apr 27) where the inference loop never returned control to the terminal pump for ~109 s while a sub-agent looped.

## Root cause

`koda-cli/src/tui_handlers_inference.rs:156-161` — the post-event drain loop is **unbounded**:

```rust
Some(ui_event) = ui_rx.recv() => {
    handle_inference_ui_inline(ui_event, ...);
    // Batch-drain queued engine events to reduce redraws.
    while let Ok(extra) = ui_rx.try_recv() {   // ⚠️ no upper bound
        handle_inference_ui_inline(extra, ...);
    }
}
```

When the engine produces events faster than `handle_inference_ui_inline` can drain them (sub-agent fan-out, parallel `ToolCallReady` events, fast streaming), `try_recv()` keeps returning `Ok` because new events arrive between iterations. The `tokio::select!` never re-evaluates, starving:

1. `crossterm_events.next()` — keystrokes, mouse events, Ctrl+C
2. `&mut turn` — the inference future itself

The existing `biased` directive (added in the original #540 fix) only prioritises arms at the select! point. It cannot help once one arm is monopolising the executor.

## Why mouse events leak as escape codes

With the event loop blocked, the OS terminal buffers raw bytes (mouse reports begin with `^[[<` under SGR mouse mode). When the loop finally pumps, crossterm's parser may mis-frame partial sequences as individual key events. Those key events then flow into the textarea, producing the visible `[<64;3;7M` garbage.

## Proposed fix

Bound the drain loop to a sane cap (e.g. 64 events per outer iteration). Yield back to the select after the cap so terminal events get a turn:

```rust
Some(ui_event) = ui_rx.recv() => {
    handle_inference_ui_inline(ui_event, ...);
    // Bounded drain — yield back to the select after N events so
    // terminal input + Ctrl+C aren't starved by event floods.
    const MAX_DRAIN: usize = 64;
    for _ in 0..MAX_DRAIN {
        match ui_rx.try_recv() {
            Ok(extra) => handle_inference_ui_inline(extra, ...),
            Err(_) => break,
        }
    }
}
```

The loop body still amortises the redraw cost across up to 64 events (the original optimisation), but no longer monopolises the executor.

## Acceptance criteria

- [ ] Drain loop has an explicit upper bound; comment explains why
- [ ] Regression test: simulate a flood of `EngineEvent::TextDelta` (e.g. 10 000 events) and assert the terminal-event arm gets serviced within N iterations
- [ ] Manual smoke: scroll wheel during a sub-agent fan-out → no escape leakage in composer
- [ ] Manual smoke: Ctrl+C during a heavy streaming turn → cancel propagates within ~100 ms

## Refs

- #540 (original mouse-leakage fix; this is the regression)
- `koda-cli/src/tui_handlers_inference.rs:120-161` (the `biased` select + drain pattern)

### LINKED ISSUE #1139: TUI: round-robin fairness in inference event loop select (codex parity)
## Context

Comparative review of `../codex` (also ratatui-based, also Rust) revealed they don't have koda's input-freeze / mouse-leakage class of bugs because their event-stream uses **explicit round-robin polling** between draw events and crossterm input — not just `biased` select.

The `biased` directive in our `tokio::select!` only chooses *which arm to poll first when multiple are ready at the same select point*. It does **not** prevent one arm from monopolising the executor *between* select points. Codex doesn't rely on `biased` at all — they rotate which channel polls first on every iteration:

```rust
// codex-rs/tui/src/tui/event_stream.rs:289-310
let draw_first = self.poll_draw_first;
self.poll_draw_first = !self.poll_draw_first;   // ← flips every poll

if draw_first {
    if let Poll::Ready(event) = self.poll_draw_event(cx) { ... }
    if let Poll::Ready(event) = self.poll_crossterm_event(cx) { ... }
} else {
    if let Poll::Ready(event) = self.poll_crossterm_event(cx) { ... }
    if let Poll::Ready(event) = self.poll_draw_event(cx) { ... }
}
```

Result: input can never starve draw, draw can never starve input, even under sustained event pressure on either side.

## Goal

Apply the same rotation discipline to koda's inference-loop `tokio::select!` (`koda-cli/src/tui_handlers_inference.rs:104-160`). Stack on top of #1137 (bounded drain) so we have *both* defenses:

- **#1137** — bound the drain so one arm can't monopolize per-iteration
- **this issue** — rotate priority across iterations so no arm consistently wins

## Proposed shape

Track an alternating flag on `TuiContext` (or a local `loop` counter in `run_inference_turn`). Use it to rebuild the select with rotated arm order each iteration:

```rust
let prefer_input = !prefer_input;
if prefer_input {
    tokio::select! {
        biased;
        Some(Ok(ev)) = self.crossterm_events.next() => { ... }
        Some(ui_event) = ui_rx.recv() => { ... }
        result = &mut turn => { ... }
    }
} else {
    tokio::select! {
        biased;
        Some(ui_event) = ui_rx.recv() => { ... }
        Some(Ok(ev)) = self.crossterm_events.next() => { ... }
        result = &mut turn => { ... }
    }
}
```

A cleaner alternative would be to merge the streams ahead of the select (codex-style) so the rotation lives in stream poll order, not in macro expansion. That's a bigger refactor — file as a follow-up if the simpler version proves brittle.

## Acceptance criteria

- [ ] No single arm of the inference-loop select can win >N consecutive iterations under load
- [ ] Regression test: simulate sustained `EngineEvent::TextDelta` flood and assert keystrokes interleave within a small bound
- [ ] No regression in throughput (drain batching still works)

## Refs

- Parent epic: TBD (umbrella issue)
- #1137 — bounded drain (must land first or together)
- codex-rs/tui/src/tui/event_stream.rs:280-315 — rotation pattern

lijunzh:
## Summary

Two defenses-in-depth on the inference event loop, ported from a comparative review of `codex`'s TUI architecture. **First PR under epic #1141** (TUI responsiveness & terminal hygiene — v0.3.x graduation blocker).

## What was wrong

The inference inner loop's `tokio::select!` had two compounding problems that caused multi-second composer freezes, mouse escape sequences leaking into the prompt, and Ctrl+C lag during heavy inference:

1. **Unbounded drain** — `while let Ok(extra) = ui_rx.try_recv() { ... }`. Under sustained event pressure, new events arrived between iterations of the inner loop, so `try_recv()` kept returning `Ok` indefinitely. The parent select never re-evaluated and terminal input was starved.

2. **`biased` is not enough** — it only chooses priority *within* one select point. Once the unbounded drain monopolised the executor, the input arm never got to run.

### 1. Bound the drain at `MAX_DRAIN = 64` events per iteration (#1137)

Extracted into a small testable helper `drain_bounded<T, F>` so the core regression guard can be unit-tested without spinning up a full `TuiContext`.

The drain still amortises N events into 1 redraw (the original optimisation that prompted the unbounded drain in the first place); it just yields back to the select after at most 64 events.

### 2. Round-robin priority between input and ui arms (#1139)

`prefer_input` flips every iteration. Two parallel `tokio::select!` expansions encode the rotation (one with input first, one with ui first), with the arm bodies factored through a `SelectArm` enum so neither block duplicates the (large) handler bodies.

Combined with #1137, this guarantees that keystrokes / Ctrl+C / scroll events get serviced within a small bounded number of inner-loop iterations under any event load.

## Tests

5 new unit tests in `drain_tests`:

- `drain_returns_zero_on_empty_channel`
- `drain_processes_all_when_below_cap`
- **`drain_stops_at_cap_and_leaves_remainder`** ← the core #1137 regression guard (sends 1000 items, asserts drain returns exactly 64, asserts the remaining 936 are still queued for the next iteration)
- `drain_with_cap_zero_processes_nothing` (defensive)
- `sustained_flood_is_drained_across_multiple_iterations` (mimics the real scenario: producer keeps refilling, drain converges across ~8 iterations at cap 64 / total 500)

Full koda-cli lib suite: 485 passed. Workspace: green. clippy: clean. fmt: clean.

## Closes

- Closes #1137
- Closes #1139
- Tracked under epic #1141

## Follow-ups in the same epic (not in this PR)

- #1138 — frame coalescing scheduler (the bigger CPU win)
- #1140 — drop `EnableMouseCapture`, use `?1007h` (eliminates the mouse-leak bug class by construction)
- #1116 — long-term hand-rolled textarea

### DIFF
--- a/koda-cli/src/tui_handlers_inference.rs
+++ b/koda-cli/src/tui_handlers_inference.rs
@@ -22,6 +22,47 @@ use ratatui::{
 };
 use tokio::sync::mpsc;
 
+/// Outcome of one inner-loop `tokio::select!` poll inside `run_inference_turn`.
+///
+/// Carrying the raw event out of the select arm (rather than handling it
+/// inline) lets us rotate the priority order of the input vs ui arms each
+/// iteration without duplicating the (large) handler bodies (#1137, #1139).
+enum SelectArm {
+    Crossterm(Event),
+    Ui(UiEvent),
+    TurnDone(anyhow::Result<()>),
+}
+
+/// Drain up to `cap` items from an `mpsc::UnboundedReceiver` synchronously,
+/// invoking `on_event` for each.
+///
+/// Returns the number of items processed (always `<= cap`).
+///
+/// This is the kernel of the #1137 fix: the original drain loop was
+/// `while let Ok(extra) = rx.try_recv() { ... }` which is unbounded. Under
+/// sustained event pressure (sub-agent fan-out, fast streaming) new events
+/// arrived between iterations of the loop, so `try_recv()` kept returning
+/// `Ok` indefinitely — the parent `tokio::select!` never re-evaluated and
+/// terminal input was starved. Bounding the drain forces a yield back to
+/// the select after a bounded number of events.
+fn drain_bounded<T, F: FnMut(T)>(
+    rx: &mut mpsc::UnboundedReceiver<T>,
+    cap: usize,
+    mut on_event: F,
+) -> usize {
+    let mut n = 0;
+    for _ in 0..cap {
+        match rx.try_recv() {
+            Ok(item) => {
+                on_event(item);
+                n += 1;
+            }
+            Err(_) => break,
+        }
+    }
+    n
+}
+
 impl TuiContext {
     /// Run a full inference turn: start the turn future, handle events
     /// inside the inner `tokio::select!` loop, and perform post-turn
@@ -48,6 +89,13 @@ impl TuiContext {
                 .session
                 .run_turn(&self.config, pending_images, &cli_sink, cmd_rx);
             tokio::pin!(turn);
+
+            // Rotate which select arm is preferred each iteration so neither
+            // terminal input nor engine events can monopolise the executor
+            // under sustained load (#1137, #1139). `biased` alone only chooses
+            // priority *within* one select point — it does not prevent one arm
+            // from winning N times in a row when both are constantly ready.
+            let mut prefer_input = true;
 
             loop {
                 // Clamp scroll offset before drawing (resize may have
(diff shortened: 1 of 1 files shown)


lijunzh:
📚 **Stack:** PR #1143 (frame coalescing, #1138) is built on top of this branch and ready for review once this lands. See epic #1141 for the broader plan.

Issue #1201, opened 2026-05-03
Bg-agent UX: WaitTask blocks parent with no live activity feedback (compare codex/claude/gemini)


lijunzh:
## Problem

When the model delegates to a bg agent and then calls `WaitTask` to gather the result, the user experiences:

1. A single `WaitTasks` tool-header line in the scrollback
2. The status pill ticks `🌀 1 agent` (just a count)
3. **Total silence** for up to 300 seconds while the sub-agent runs
4. The main agent appears completely blocked — because it actually is

The parent inference loop `await`s on the WaitTask call. No streaming, no thoughts, no tool-call breadcrumbs from the child surface in the parent's TUI. The user has no signal whether the sub-agent is making progress, stuck, or about to time out.

## What we already have (and are wasting)

`BgStatusEmitter` (`koda-core/src/bg_agent.rs:286`) already pushes per-iteration heartbeats:
```rust
EngineEvent::BgTaskUpdate { task_id, spawner, status: AgentStatus::Running { iter } }
```
But the TUI renderer (`tui_render.rs:287`) explicitly drops these on the floor — they only update the count in the status pill. Per-tool activity from inside the bg agent is never emitted, so even if we wired the events up, we'd only show "iter 3 → iter 4" — useless.

## How the competitors handle it

I read source for codex, claude_code, and gemini-cli. Three distinct philosophies:

### 🟢 **Gemini CLI** — inline streaming activity feed
`packages/cli/src/ui/components/messages/SubagentProgressDisplay.tsx`

Renders the parent's view with a live block:
```
Running subagent foo...
  💭 Looking at the auth module to find the bug...
  ⠋ Read   src/auth/login.rs
  ✓ Read   src/auth/session.rs
  ⠋ Bash   cargo test auth::
```

Backed by streaming `SubagentActivityEvent`s (`packages/core/src/agents/types.ts:84`):
- `TOOL_CALL_START` / `TOOL_CALL_END` → tool-call entries with status (spinner→✓→✗→⊘)
- `THOUGHT_CHUNK` → 💭 italic reasoning lines
- `ERROR` → red error rows

`SubagentProgress.recentActivity: SubagentActivityItem[]` — each item has type, displayName, args, status. Args truncated at 60 chars. Result rendered as markdown when done.

**Verdict: this is the gold standard for our "in-place" UX. We could implement most of it on top of our existing `BgStatusEmitter`.**

### 🟡 **Claude Code** — per-agent progress line
[...]

PR #1207, opened 2026-05-03
Drop redundant per-event BgChildActivity TUI render (#1201 B2)

lijunzh:
Polish follow-up to #1206 (B1). Closes the **B2 bullet** of #1201.

## TL;DR

B1 made bg-agent activity *visible* but, in TUI, rendered it twice:
- **Live** \u2014 50 dim lines as events streamed in
- **Post-completion** \u2014 the same 50 events as bullets under `\u2705 Background agent X completed` (already there since #1022 B9)

Result: a 50-tool bg agent dumped **100 dim lines + the trace** into scrollback. B2 keeps the durable record and drops the live duplicate.

## What changed

| File | Lines | What |
|---|---|---|
| `koda-cli/src/tui_render.rs` | -23 / +27 | `BgChildActivity` arm folds into the no-op batch with `BgTaskUpdate` / `TodoUpdate`. Long comment block explains *why* so a future refactor doesn't re-introduce the noise |

**Headless and ACP are untouched.** Tail-following contexts (`tee`, `grep`, IDEs) genuinely benefit from per-event streaming \u2014 different ergonomic.

## Why not a Gemini-style live block?

Considered. Three scopes were on the table:

| Option | Effort | Picked? |
|---|---|---|
| **A** Drop TUI per-event render | ~50 LOC | \u2705 |
| **B** In-place updating line per task | ~250 LOC + `ScrollBuffer::replace` API | \u274c |
| **C** Footer pad above composer + collapse-on-completion | ~500 LOC + viewport surgery | \u274c |

Picked **A** because:
- It solves the *actual* user pain (noise)
- The post-completion \u2705 trace is already a complete record
- Status pill (`X bg agents running`) + `/agents` panel cover "is it alive"
- Respects YAGNI \u2014 if users want live overlay, B3 can build it on the same `BgChildActivity` event already plumbed in #1206

## Coverage

| Surface | Live per-event | Post-completion summary |
|---|---|---|
| TUI | \u274c (this PR) | \u2705 `\u2705 Background agent X completed\n<trace>` |
| Headless (`koda --no-tui`) | \u2705 `[bg N] \ud83d\udd27 Read foo.rs` via stderr | \u2705 |
| ACP (IDE) | \u2705 `[bg task N] Read foo.rs` text-chunk | \u2705 |

## Tests

- **+1 test**: `bg_child_activity_does_not_push_lines_to_scroll_buffer` pins the no-op for all three `BgChildActivityKind` variants (ToolStart / ToolEnd / Info)
- 592/592 koda-cli tests pass (was 591)

## Verification

```
cargo test -p koda-cli --lib                            \u2192 592/592
cargo clippy --workspace --all-targets ... -D warnings  \u2192 clean
cargo fmt --all -- --check                              \u2192 clean
python3 scripts/check_tokio_test_flavor.py              \u2192 OK
```

## Closes

The **B2 bullet** of #1201. A2 / Bonus / B3 (live overlay) remain open in #1201.

### DIFF
--- a/koda-cli/src/tui_render.rs
+++ b/koda-cli/src/tui_render.rs
@@ -285,7 +285,8 @@ impl TuiRenderer {
             | EngineEvent::TurnEnd { .. }
             | EngineEvent::LoopCapReached { .. }
             | EngineEvent::BgTaskUpdate { .. }
-            | EngineEvent::TodoUpdate { .. } => {
+            | EngineEvent::TodoUpdate { .. }
+            | EngineEvent::BgChildActivity { .. } => {
                 // Handled by the event loop, not the renderer.
                 // (#1076) BgTaskUpdate is routed to the bg-task panel,
                 // which still reads from the registry snapshot for its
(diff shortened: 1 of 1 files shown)


lijunzh:
## Sequencing note: prerequisite for #1210

Field study (#1210) confirms this PR is the **right** intermediate state — it removes the duplicate per-event render that conflicts with the live overlay widget planned in #1210.

Without #1207, the #1210 overlay would render activity **three times**: dim live lines + overlay + post-completion bullets. With #1207, the scroll buffer stays clean during wait, the overlay handles live signal, and the post-completion bullets remain as the durable record.

Brief silent-during-wait window is acceptable — #1210 lands immediately after to fill it. Merging.

PR #1322, opened 2026-05-06
Fix #1321: pump bg-agent status events to sink during tool dispatch

### LINKED ISSUE #1321: UX audit: silent slash-command failures + WaitTask UI silence
# UX audit: silent slash-command failures + WaitTask UI silence

## Background

Triggered by analysis of debug bundle `koda-debug-20260506-225301-session.zip` (user running koda **0.3.0**), where the user reported:

> when wait on sub agent, there is no UI interaction. after WaitTask completes, the model was telling it is waiting on results. Was I imagining things?
> Also: `/export` was used but no message that it is not a valid command — should we just fail silently?

Both are real bugs. This issue documents what's confirmed, what's already partially fixed in `main`, and what's still wrong.

---

### Repro

1. Type `/export` (or `/exprt`, or `/agents`, or any non-existent slash) at idle.
2. Press Enter.

### Expected

Either:
- Error line: `✗ Unknown command: /export. Type /help to list commands.` (gemini-cli pattern), or
- Suggestion: `✗ Unknown /export. Did you mean /expand?`, or
- Treat as user message and forward to model (codex's `ParsedCommand::Unknown` path).

### Actual

Input is silently swallowed. No error, no warning, nothing in scrollback. The user reasonably assumes the command ran. This is a textbook violation of the principle "errors should never pass silently" (Zen of Python).

### Root cause

[`koda-cli/src/tui_commands.rs:268`](koda-cli/src/tui_commands.rs#L268):

```rust
match cmd {
    "/exit" => SlashAction::Quit,
    "/model" => match arg { ... },
    // ... 17 more named arms ...
    "/mcp" => { dispatch_mcp(...).await; SlashAction::Continue }

    _ => SlashAction::Continue,        // ← swallows everything unknown
}
```

`Continue` then triggers `reinit_after_slash()` and returns `Handled` to the caller — which records the input as a successfully-handled command and clears the textarea. Zero user-visible feedback.

### Affected commands (real examples)

- `/export` — never existed in this codebase (referenced in a stale comment in `wait_task_format.rs:11`)
- `/agents` (plural) — **deleted in #1210** in favor of the always-on overlay; no migration warning
- `/keys` — handled (`"/key" | "/keys"` arm, line 247) ✓ but only because someone remembered
- Any typo like `/expnd`, `/comapct`, `/halp`

### Bonus — registry / handler drift

`SLASH_COMMANDS` (`composer/slash_popup.rs:43`) advertises 20 commands. `handle_slash_command` has 19 named arms, plus `/vim`, `/model`, `/provider`, `/sessions` special-cased earlier in `dispatch_slash` (`tui_context/mod.rs:644`). The lists are kept in sync **manually**. There's no compile-time check, so any new command added to one place but not the other will silently no-op.

### Suggested fix (one of)

**A. Loud-fail (simplest, gemini-cli pattern):**

```rust
_ => {
    tui_output::err_msg(
        buffer,
        format!("Unknown command: {cmd}. Type /help to list commands."),
    );
    SlashAction::Continue
}
```

**B. Loud-fail with did-you-mean** (use `strsim::jaro_winkler` against `SLASH_COMMANDS` keys, suggest closest match if score > 0.85). ~10 LOC.

**C. Single source of truth**: make `SLASH_COMMANDS` an enum-with-handlers and dispatch via match-on-enum so the compiler enforces parity. Bigger refactor; folds B as a follow-up.

Recommend **B** — small, kind, matches the codex/gemini-cli reference. C is a good follow-up.

---

### Repro (as captured in debug bundle, koda 0.3.0)

```
22:49:20  user: "quick explore the codebase by sub agents..."
22:49:20  → InvokeAgent(explore)              ← spawned (agent:1)
22:49:21  → text: "I'm waiting for the exploration agent..."
22:49:21  → WaitTask(["agent:1"])             ← BLOCKS HERE
22:50:21  ← timed_out (60s of UI silence)     ← user sees nothing for 60s
22:50:23  → text: "taking a bit longer..."
22:50:23  → WaitTask(["agent:1"], 120)
22:52:23  ← timed_out (120s of UI silence)    ← another 120s of silence
22:52:26  → text: "still working..."
22:52:26  → WaitTask(["agent:1"], 120)
22:52:31  ← completed (5s)
22:52:34  → final summary
```

The "I'm waiting…" / "taking a bit longer…" lines all rendered **after** each `WaitTask` returned, not during. So between the `→` and `←` lines: dead silent TUI. User can't tell if koda is alive, the agent is alive, or anything is making progress. **The user wasn't imagining it.**

### Why this happened on 0.3.0

The user's binary is **koda 0.3.0**. None of these mitigations are in 0.3.0:

- #1210 — `widgets::child_activity_overlay` (live activity row above composer)
- #1233/#1234 — `ChildAgentRegistry` rename + activity tap point in `ui_handler`
- #1163 (just merged, will be 0.3.3) — `ForwardingChildSink` so child events actually flow to the parent's UI in real-time

So on 0.3.0 the user genuinely had a frozen TUI for 5 minutes. Recommend a 0.3.3 release note that calls this out.

### What's still imperfect on `main`

Even with all of #1210/#1234/#1163 in place, gaps remain:

1. **Activity overlay only updates on `ChildAgentActivity` events.** If the bg agent is itself stuck inside a slow provider call (e.g., Gemini taking 30s to think), no `ToolStart` events fire — so the overlay's "age" badge silently grows stale ("12s") with no indicator that the engine is fine, just waiting.
   - Fix idea: tick the overlay's `frame_requester` on a 1Hz timer when ≥1 row is `Running` so the `(age)` column counts up live, even without new events.
   - There is a per-iteration `Running { iter: N }` `ChildTaskUpdate` heartbeat (`engine/event.rs:168`) — confirm the overlay actually consumes it; if not, that's the wire-up bug.

2. **The scrollback `WaitTask` header has no progress affordance.** Once printed, it just sits there. Compare with codex's tool-call rendering where in-progress tools show an animated indicator. The `streaming_tool_ids` machinery in `tui_render.rs` already exists for tools that emit `ToolOutputLine` — `WaitTask` could synthesize one such line per completed iter heartbeat ("agent:1 → iter 3, last: Read src/foo.rs") without flooding scrollback.

3. **The model's narration ("I'm waiting…") is informational fluff that costs tokens.** It only exists because the model has no other way to signal "I haven't crashed". A live overlay is strictly better for the user, and the system prompt could encourage shorter narration when the overlay covers it.

### Suggested fix scope

- **Sub-task A** (small): wire 1Hz timer-driven redraw when overlay has `Running` rows, so age counts up.
- **Sub-task B** (small): verify `ChildTaskUpdate::Running { iter: N }` reaches the overlay tracker; if not, plumb it.
- **Sub-task C** (medium): synthesize `WaitTask` scrollback progress lines from forwarded child activity, gated behind `/verbose` so non-verbose users keep clean scroll.
- **Sub-task D** (doc): release-notes entry for 0.3.3 calling out the WaitTask UX overhaul stack (#1163 + #1210).

---

### B1. `/agents` quietly removed without deprecation breadcrumb

`#1210` deleted the `/agents` interactive panel in favor of the overlay. Anyone with muscle memory typing `/agents` now hits Bug #1's silent consumption. After the Bug #1 fix lands, `/agents` will show "Unknown command" — but a friendlier path is a one-line did-you-mean stub:

```rust
"/agents" => {
    tui_output::info_msg(buffer, "/agents was replaced by the always-on activity overlay above the composer.".into());
    SlashAction::Continue
}
```

### B2. Slash-popup, help-output, and dispatch table are three separate hand-maintained lists

The friction point that breeds Bug #1 in the first place. Pull this out as its own refactor ticket (Suggested fix C above).

### B3. Mid-inference slash guard already does this right (#1211)

When user types `/cancel` mid-turn, `is_slash_command_attempt` rejects with a visible warning ([`crossterm_handler.rs:322`](koda-cli/src/inference/crossterm_handler.rs#L322)). This is the exact pattern Bug #1 needs at idle — the prior art is in-repo.

### B4. `wait_task_format.rs:11` references `/export` as a real-but-renamed command

> "until RFC #1167 PR δ collapsed `/export` into `/debug-bundle`"

This explains where the user got the idea `/export` was a thing — it WAS a real command that got renamed, but the comment is the only surviving reference. Worth either restoring `/export` as an alias for `/debug-bundle` (Postel: be liberal in what you accept) or scrubbing the stale comment.

---

## Reference comparison (per repo conventions)

| Surface | codex | claude_code | gemini-cli | koda (current) |
|---|---|---|---|---|
| Unknown slash | `ParsedCommand::Unknown` → forwarded to model | error line | "Unknown command: X" error | **silent** ❌ |
| Did-you-mean | no | no | no | could be 1st 🐶 |
| Live bg-agent activity | summary row in tool block | top-of-screen status | inline tool log | overlay (#1210, post-0.3.0) ✓ |
| Heartbeat during slow LLM call | n/a (no bg agents) | n/a | n/a | partial (event-driven, no timer) |

---

## Acceptance criteria for first PR (Bug #1 only)

- [ ] Typing `/anything-not-in-the-table` shows `✗ Unknown command: /anything-not-in-the-table. Type /help.`
- [ ] If a close match exists (≥0.85 jaro_winkler), append "  Did you mean `/closest`?"
- [ ] Existing valid slash commands continue to work
- [ ] New regression test in `tui_commands.rs` — unknown command produces an `err_msg` line in the buffer
- [ ] `/export` and `/agents` get optional friendly-redirect arms (B1, B4) — or punted to a follow-up

Bug #2 sub-tasks A–D get their own PRs since they touch different layers.

---

🐶 Generated by code-puppy from analysis of `~/.config/koda/debug-bundles/koda-debug-20260506-225301-session.zip` against current `main` (post-#1163).

lijunzh:
Fixes #1321.

## TL;DR

Bg-agent activity events were correctly produced, correctly enqueued, correctly forwarded by every layer of the engine — and then quietly buffered in a `VecDeque` for the entire duration of any long-running tool call (notably `WaitTask`, up to 300s). The activity overlay went dark until the wait returned, at which point a flood of stale events landed in a single frame. This PR adds a 200ms drain pump that runs alongside tool dispatch so events flow live.

## The pipeline (with the missing piece)

```
bg agent emits ToolCallStart
 → ForwardingChildSink::emit                                  ✅ #1163
   → emitter.send_activity                                    ✅
     → registry.push_status_event                             ✅
       → events.lock().push_back(event)   ← VecDeque<EngineEvent>
                                            💀 only drained at the top
                                            💀 of the next inference iteration
                                            💀 (inference.rs:741)
       ↓
   parent's EngineSink                                        ❌ silent during WaitTask
     → activity overlay                                       ❌ silent during WaitTask
```

Production callers of `drain_status_events`: **exactly one**, between parent inference iterations. So while the parent is blocked inside `WaitTask`, nobody pumps the queue. This PR adds a second consumer that runs *during* tool dispatch.

## What this PR does

1. **New module `koda-core/src/status_pump.rs`** with `with_status_pump(fut, bg_agents, sink)` — wraps any future and races it against a periodic drain task (`STATUS_PUMP_INTERVAL = 200ms`). On future completion, a final flush catches anything that landed in the last tick window.

2. **Wires the helper around all three `execute_tools_*` arms** in `inference_loop` (`koda-core/src/inference.rs`):
   - `execute_tools_parallel`
   - `execute_tools_split_batch`
   - `execute_tools_sequential`

   So the fix covers every dispatch shape and generalizes to *any* slow tool, not just `WaitTask`.

3. **Four new tests** in `status_pump::tests` that pin the bug-and-fix contract — see "Bug-pinning" section below.

## Why this shape (vs the alternatives)

The issue body proposed three fix options. Picked Option 3 (slightly evolved) for blast-radius reasons:

| Option | Pro | Con |
|--------|-----|-----|
| 1. Replace queue with `tokio::mpsc`, give rx to UI loop | Architecturally cleanest | Requires `InferenceContext::sink: &dyn → Arc<dyn>` and rewiring every sub-agent dispatch site |
| 2. Pump from inside `WaitTask` | Localized | Doesn't generalize to other slow tools; `WaitTask` lacks sink access |
| **3. Pump at dispatch boundary (this PR)** | Sink + registry already in scope; covers all slow tools; ~330 LOC | Adds one always-pinned future peatch call (negligible) |

## Concurrency safety

The pump and the existing per-iteration drain at `inference.rs:741` both call `ChildAgentRegistry::drain_status_events()`, which takes the registry's internal mutex and drains atomically. Concurrent callers see whichever drain wins the lock; the loser sees an empty queue. **Events are never duplicated.** The per-iteration drain stays as a backstop for events queued while no tool dispatch is in flight (e.g. between turns).

## Bug-pinning

The headline test is `pump_drains_events_during_long_running_future`:

- Producer task pushes 5 events spaced 80ms apart over ~400ms.
- Wraps a 500ms `tokio::time::sleep` in `with_status_pump`.
- Asserts all 5 events reach the sink in FIFO order before the future returns.

**Verified counterfactual**: I temporarily disabled the pump body (just the `for ev in bg_agents.drain_status_events()` block — leaving the rest of the helper intact) and reran. Result: `assertion failed: expected 5 events, got 0`. That's the *exact* pre-fix behavior — the queue fills, nobody drains, and the future returns before any events flow. Restoring the pump → 5/5 events delivered. Test pins the bug correctly.

The other three tests cover:
- `pump_final_flush_catches_events_emitted_in_last_tick_window` — events pushed in the last 200ms window before completion still land via the explicit final drain.
- `pump_is_noop_when_queue_stays_empty` — common-case overhead check (most tool calls have no concurrent bg-agent activity).
- `pump_propagates_future_errors` — `?`-style error propagation at the dispatch sites still works untouched.

## Verification

- ✅ `cargo build --workspace` clean
- ✅ `cargo test -p koda-core --lib` — **1336 passed, 0 failed**
- ✅ `cargo test -p koda-core --lib status_pump::` — 4/4 passed, ran 5x consecutively, no flakes
- ✅ `cargo clippy -p koda-core --lib --tests -- -D warnings` clean
- ✅ `cargo fmt --check -p koda-core` clean

## What this does not change

- No public API changes. `drain_status_events` and `push_status_event` keep their existing signatures.
- No sink ownership refactor. `InferenceContext::sink` stays `&dyn EngineSink`.
- No behavior change for tool calls that complete in <200ms — the pump never ticks, only the final flush runs (typically also empty).
- No behavior change between turns. The existing per-iteration drain at `inference.rs:741` still fires first thing on every new iteration.

## Manual smoke test recipe

For anyone who wants to eyeball it:

1. Build the binary: `cargo build --release`
2. Spawn a bg agent that does several slow tool calls: `Hey explore agent, read these 5 files one at a time with a 2-second think between each: foo.rs, bar.rs, baz.rs, qux.rs, quux.rs`
3. Immediately wait on it: `WaitTask agent:1 timeout 60`
4. Watch the activity overlay above the composer. Pre-fix: silent for 10+ seconds, then a flash. Post-fix: live "Read foo.rs", "Read bar.rs", … updates every couple of seconds.

🐶

### DIFF
--- a/koda-core/src/inference.rs
+++ b/koda-core/src/inference.rs
@@ -1312,14 +1312,37 @@ async fn inference_loop_inner(ctx: InferenceContext<'_>) -> Result<()> {
         // sees the same Koda-owned-file downgrade the sequential path
         // sees. Without it, batches containing `Delete owned.tmp` are
         // spuriously refused parallelization (perf regression).
+        //
+        // #1321: each dispatch arm is wrapped in `with_status_pump` so
+        // bg-agent `ChildAgentActivity` events flow to the parent's
+        // sink in real time during long tool calls (notably `WaitTask`).
+        // Pre-#1321 these events sat in the registry queue until the
+        // top of the next inference iteration, making the activity
+        // overlay go dark for the entire wait — see the module docs
+        // on `crate::status_pump` for the full story.
         if remaining_tools.len() > 1
             && can_parallelize(&remaining_tools, mode, project_root, Some(file_tracker))
         {
-            execute_tools_parallel(&remaining_tools, tool_ctx, file_tracker).await?;
+            crate::status_pump::with_status_pump(
+                execute_tools_parallel(&remaining_tools, tool_ctx, file_tracker),
+                bg_agents,
+                sink,
+            )
+            .await?;
         } else if remaining_tools.len() > 1 {
-            execute_tools_split_batch(&remaining_tools, tool_ctx, cmd_rx, file_tracker).await?;
+            crate::status_pump::with_status_pump(
+                execute_tools_split_batch(&remaining_tools, tool_ctx, cmd_rx, file_tracker),
+                bg_agents,
+                sink,
+            )
+            .await?;
         } else if !remaining_tools.is_empty() {
-            execute_tools_sequential(&remaining_tools, tool_ctx, cmd_rx, file_tracker).await?;
+            crate::status_pump::with_status_pump(
+                execute_tools_sequential(&remaining_tools, tool_ctx, cmd_rx, file_tracker),
+                bg_agents,
+                sink,
+            )
+            .await?;
         }
 
         // Update skill scope: if any ActivateSkill call was made, check whether
--- a/koda-core/src/lib.rs
+++ b/koda-core/src/lib.rs
@@ -117,6 +117,10 @@ pub mod session;
 pub mod skill_scope;
 /// Skill discovery and activation (project, user, built-in).
 pub mod skills;
+/// Status-event pump that bridges the gap between bg-agent emissions
+/// and the parent's `EngineSink` while a long-running tool call is in
+/// flight (#1321). See module docs for the full design rationale.
+pub mod status_pump;
 /// Sub-agent result caching — zero-cost retries after compaction.
 pub mod sub_agent_cache;
 /// Sub-agent invocation and lifecycle management.
--- a/koda-core/src/status_pump.rs
+++ b/koda-core/src/status_pump.rs
@@ -1,0 +1,300 @@
+//! Status-event pump for long-running tool calls (#1321).
+//!
+//! ## The problem this module fixes
+//!
+//! Background sub-agents emit `ChildAgentActivity` events through their
+//! `ForwardingChildSink` (see `engine::sink`). Those events land in
+//! [`crate::child_agent::ChildAgentRegistry`]'s status queue via
+//! `push_status_event`. Until #1321 there was **exactly one** consumer of
+//! that queue in production — the per-iteration drain at the top of
+//! [`crate::inference::inference_loop`] — so events only reached the
+//! parent's `EngineSink` *between* parent inference iterations.
+//!
+//! While the parent was blocked inside a long-running tool call (the
+//! textbook offender is `WaitTask`, which can block for up to 300s), the
+//! activity overlay above the composer received zero updates. Then, the
+//! moment the wait returned, every queued event flushed in a single frame
+//! — by which time the bg agent had already finished and the events were
+//! useless as live progress.
+//!
+//! ## The fix
+//!
+//! [`with_status_pump`] races the wrapped tool-dispatch future against
+//! a background drain task that ticks every [`STATUS_PUMP_INTERVAL`].
+//! Each tick drains the registry's status queue and emits each event
+//! through the parent's sink — so the activity overlay sees bg tool
+//! calls in real time, not in a stale post-wait flood.
+//!
+//! ## Why this shape
+//!
+//! Three options were considered (see #1321 issue body):
+//!
+//! 1. **Replace queue with `tokio::mpsc`**, hand the receiver to the UI
+//!    loop. Architecturally cleanest but requires changing
+//!    `InferenceContext::sink` from `&dyn EngineSink` to `Arc<dyn
+//!    EngineSink>` and wiring a new mpsc through every sub-agent
+//!    dispatch site — large blast radius for a focused bug.
+//!
+//! 2. **Pump from inside `WaitTask`**. Localized but doesn't generalize
+//!    to other slow tools (a long `Bash` or a slow MCP call has the
+//!    same problem in principle), and `WaitTask` doesn't have direct
+//!    access to the parent's `EngineSink`.
+//!
+//! 3. **Pump at the dispatch boundary** (this module). The sink and
+//!    registry are both already in scope at the three
+//!    `execute_tools_*` call sites in `inference_loop`. Wrapping each
+//!    of those three calls covers every long tool, sequential or
+//!    parallel. Smallest possible diff that solves the root cause.
+//!
+//! ## Concurrency model
+//!
+//! `drain_status_events` takes the registry's mutex and atomically
+//! drains. Two concurrent drainers (the pump and the per-iteration
+//! drain in `inference_loop`) cannot double-emit because the queue is
+//! drained inside the lock — whichever wins the race takes everything
+//! and the loser sees an empty drain. The per-iteration drain stays as
+//! a backstop for events queued while no tool dispatch is in flight
+//! (e.g. between turns).
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use crate::child_agent::ChildAgentRegistry;
+use crate::engine::EngineSink;
+
+/// How often the pump drains the status queue while a tool call is in
+/// flight. 200ms balances "smooth-feeling live overlay" against pump
+/// overhead (one mutex acquisition + a `VecDeque::drain` per tick — a
+/// few microseconds at most).
+///
+/// At 200ms the activity overlay updates ~5 times per second, which
+/// is comfortably under the human flicker-fusion threshold (~10 Hz)
+/// while staying well clear of the rate-limiter inside the TUI's
+/// frame scheduler (~120 FPS cap).
+pub const STATUS_PUMP_INTERVAL: Duration = Duration::from_millis(200);
+
+/// Run `fut` while continuously draining `bg_agents.drain_status_events()`
+/// to `sink` every [`STATUS_PUMP_INTERVAL`]. Returns `fut`'s output.
+///
+/// **Final flush guarantee**: any events emitted in the final tick
+/// window — between the pump's last drain and `fut`'s completion —
+/// are flushed to the sink before this function returns. Without this
+/// the user could miss the bg agent's terminal `ToolEnd` event when
+/// `fut` is `WaitTask` and the bg agent finishes in the same window.
+///
+/// **Safety against double-emission**: the pump and the
+/// inference-loop's per-iteration drain both call
+/// [`ChildAgentRegistry::drain_status_events`], which takes the
+/// registry's internal mutex and drains atomically. Concurrent
+/// callers see whichever drain wins the lock; the loser sees an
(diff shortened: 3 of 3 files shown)


lijunzh:
Closing in favor of #1323, which lands the proper architectural fix.

## Why close this

#1322 added a 200 ms `tokio::time::interval` pump task next to the existing `Mutex<VecDeque>` queue + drainer pattern — a backstop that bridged the gap during long tool calls without changing the underlying shape. It was a sound hotfix and would have shipped a working fix, but it cemented a design that neither reference implementation uses.

#1323 replaces the queue with an `mpsc::UnboundedSender` and a session-spawned forwarder task that pulls events continuously, mirroring [Codex's `forward_events`](https://github.com/openai/codex/blob/main/codex-rs/core/src/codex_delegate.rs) and Claude Code's `for await` over `runAgent()`. The pump becomes structurally unnecessary because there is no longer a queue for events to sit in.

| | #1322 (this PR — pump hotfix) | #1323 (mpsc + forwarder) |
|--|--|--|
| Pattern | Queue + drainer + 200 ms pump | Channel + per-session forwarder |
| Latency during tool call | ≤200 ms | live (~µs) |
| Matches Codex / Claude Code shape | No | Yes |
| Net diff | +330 LOC | +306 / −35 LOC |

The reference audit in #1321 walks through both reference designs in detail.

No code from #1322 is being lost — every problem it identified is solved by #1323, and the inline-doc rationale here was useful enough that it shaped the doc comments on `attach_event_sink` and the new `forwarder_emits_during_blocking_consumer` test.

Thanks for being the hotfix that made the shape obvious. 🐶

Issue #1327, opened 2026-05-07
Flaky: safe_trust_sub_agent_can_write_files races under workspace-parallel CI load


lijunzh:
## Symptom

`koda-core/tests/e2e_sub_agent_trust_test.rs::safe_trust_sub_agent_can_write_files` flakes under workspace-parallel CI load, panicking with:

```
thread 'safe_trust_sub_agent_can_write_files' panicked at koda-core/tests/e2e_sub_agent_trust_test.rs:105:5:
sub-agent's Write must produce the file at "/tmp/.tmpXXXXXX/sub_agent_wrote_this.txt"
(pre-#1250 this would have auto-rejected with no-channel error)
```

Real-world hit: PR #1326 CI run `25471442843` (Test job ID `74735985607`). The failure was masked by #1326-companion-issue's `pipefail` bug (the Test step appeared green and the bwrap assertion took the blame), which made it 10× harder to diagnose. After a `--failed` rerun the same test passed.

## History

This test already has a deflake attempt baked in (lines 89–98 of the test source):

```rust
// **#1321/#1323 deflake**: this test was racing parent-loop
// termination against bg-agent completion (the parent's Mock
// provider returns Text after InvokeAgent, terminating the loop
// before the bg agent's Write tool dispatches). Pre-deflake this
// failed roughly 4/5 on `main` and 1/3 on the #1323 branch. Wait
// explicitly for the bg agent to reach a terminal state — same
// pattern every other bg-agent e2e test in this directory uses.
let _bg_events = env
    .collect_bg_events_after(events.clone(), std::time::Duration::from_secs(10))
    .await
    .expect("bg sub-agent never reached terminal state");
```

The `collect_bg_events_after` wait reduced the flake rate enough to merge #1321/#1323, but it's not zero. Single-test isolated runs are clean (10/10 on main, 10/10 on PR #1326 HEAD), but `cargo test --workspace --features koda-core/test-support` runs many test binaries in parallel and the contention is enough to re-surface the race.

## What we know

1. **The bg sub-agent reaches a terminal state.** `collect_bg_events_after` returns successfully — otherwise the `.expect()` would fire instead of the `target_path.exists()` assertion. So it's not a "sub-agent never started" race.
2. **The terminal state isn't "wrote the file successfully".** The `assert!(target_path.exists())` at line 105 is the actual panic site, meaning the sub-agent terminated *without* successfully completing its Write tool call.
[...]

Issue #1104, opened 2026-04-28
Bundle: v0.2.22 release-time polish (P3 audit findings)


lijunzh:
Tracking issue consolidating low-severity items flagged by AI auditor agents during the v0.2.22 release prep (PR #1103).

Bundled per the koda-release skill's audit-dust prevention rule: N separate issues for tiny doc / metadata fixes is tracker noise that doesn't get triaged. Each item below was explicitly classified P3 by the agent that filed it (or downgraded to P3 during triage by the bullet-point heuristics: "current code is safe", "purely additive", "cosmetic", "speculative tooling without driver").

---

## Status (PR #1115 landed 5/7)

| # | Item | Status |
|---|---|---|
| 1 | `rust-version` MSRV | ✅ shipped (PR #1115) — pinned to **1.87** (auditor said 1.85, but real code uses `is_multiple_of` from 1.87) |
| 2 | `koda-core` categories | ✅ shipped (PR #1115) |
| 3 | `cargo update` | ❌ deferred — `libc 0.2.186` + `rpassword 7.5.0` is build-broken on macOS (upstream rpassword bug). See PR #1115 receipts. |
| 4 | `ratatui-textarea` bump | ❌ deferred (no upgrade target — see corrected item below). Strategic follow-up filed: #1116 |
| 5 | TOCTOU in `apply_git_config_deny` | ✅ shipped (PR #1115) |
| 6 | Worktree handling in `apply_git_config_deny` | ✅ shipped (PR #1115) |
| 7 | CI guard for SEC-002 regression tests | ✅ shipped (PR #1115) |

---

### Items

- [x] **Add `rust-version` (MSRV) to `[workspace.package]`** — flagged by rust-programmer. Workspace currently declares no MSRV. Recommend `rust-version = "1.85"` (matches `edition = "2024"` minimum) so downstream consumers get a clear MSRV signal rather than mysterious build failures on older toolchains.
  - **Resolution (PR #1115):** Pinned to `1.87`, not 1.85. The codebase already uses `u64::is_multiple_of` (stable in 1.87) so 1.85 would have been a lie. Bonus: cargo's MSRV-aware resolver now blocks 6 deps that would have required Rust 1.88+ during the next `cargo update`.

[...]

PR #1331, opened 2026-05-07
fix(ci): add pipefail to Test step + port tokio-flavor guard to bash (#1328)

### LINKED ISSUE #1328: CI: Test step swallows cargo's exit code, blames bwrap for unrelated failures
## Symptom

When `cargo test --workspace` fails inside the `Test` step in `.github/workflows/ci.yml`, the step is reported as **passing** (✓) and the very next step — `Assert bwrap SEC-002 regression tests ran (#1104)` — fails with the misleading message:

```
::error::Required bwrap regression tests did not run:
  - git_config_deny_pre_creates_for_non_git_dir_to_close_toctou
  - git_config_deny_adds_ro_bind_for_existing_git_dir
  - git_config_deny_skips_worktree_gitdir_pointer
  - git_config_deny_preserves_existing_config_mtime
```

bwrap had nothing to do with it. The root cause was a different test failing in `koda-core`, which made cargo bail out before `koda-sandbox`'s bwrap unittests finished running, leaving the bwrap test names absent from `test-output.log`.

Real-world example: PR #1326 CI run `25471442843`, where `e2e_sub_agent_trust_test::safe_trust_sub_agent_can_write_files` failed (a known flake) and the bwrap assertion took the blame. Cost ~30 min of debugging to figure out the bwrap message was a lie. See #1327 for the flake itself.

## Root cause

In `.github/workflows/ci.yml`:

```yaml
- name: Test
  run: cargo test --workspace --features koda-core/test-support 2>&1 | tee test-output.log
```

GitHub Actions runs steps with `bash -e {0}` by default — `errexit` is on, but **`pipefail` is not**. In a pipeline, only the last command's exit code is observed. `tee` always exits 0, so any non-zero from `cargo test` is silently swallowed.

This means:

1. **Real test failures look like passing steps.** Only the downstream assertion guards (bwrap, SEC-002, security-boundary) catch them, and only when their tests happen to be in the "didn't print yet" bucket when cargo bailed.
2. **The wrong step gets blamed.** A racy unrelated test failure shows up in CI as "bwrap regression tests didn't run", which sends the on-call down a security-critical rabbit hole that has nothing to do with the actual bug.
3. **The blame surface is non-deterministic.** Whether the bwrap assertion or the SEC-002 assertion fires depends on cargo's parallel scheduling at the moment of failure.

## Proposed fix

One line. Add `set -o pipefail` (or use `${PIPESTATUS[0]}`) so the step's exit code reflects cargo, not tee:

```yaml
- name: Test
  shell: bash
  run: |
    set -o pipefail
    cargo test --workspace --features koda-core/test-support 2>&1 | tee test-output.log
```

After this fix, a real cargo test failure surfaces as `Test (ubuntu-latest)` failing with the actual test name in the annotations — and the bwrap assertion step is skipped (because `errexit` cuts the job at the failed step), so it stops generating false alarms.

## Why this is worth fixing now

The recovery path for a future on-call seeing `Required bwrap regression tests did not run` is currently:

1. Read the bwrap assertion source to learn what tests it expects.
2. Notice they don't appear in the GHA log.
3. Suspect Linux toolchain, bubblewrap install, AppArmor gating, etc.
4. Eventually scroll up far enough in the log to find a `test result: FAILED` line for an unrelated test.
5. Realize bwrap was a red herring.

That's a lot of cognitive load for a problem solved by 19 characters of bash.

## Acceptance criteria

- [ ] `Test` step exits non-zero whenever `cargo test` exits non-zero.
- [ ] On a synthetic test failure, the GHA annotations point at the actual failing test, not at bwrap.
- [ ] Existing passing-CI behavior unchanged (regression check via merging into a PR with all-passing tests).

## References

- Discovered in: PR #1326 CI run https://github.com/lijunzh/koda/actions/runs/25471442843
- Related flaky-test issue: #1327
- Bwrap assertion source: `.github/workflows/ci.yml`, "Assert bwrap SEC-002 regression tests ran (#1104)" step

lijunzh:
## Summary

Closes #1328 — the missing-pipefail bug that's been masking real test failures behind misleading "bwrap regression tests didn't run" errors. While in the workflow file, also did the user-requested follow-up of porting `scripts/check_tokio_test_flavor.py` to bash, removing the one stray Python dependency from an otherwise-pure-Rust workspace.

## The bug (#1328)

The Test step in `.github/workflows/ci.yml` runs:

```yaml
- name: Test
  run: cargo test --workspace --features koda-core/test-support 2>&1 | tee test-output.log
```

GH Actions defaults `run:` blocks to `bash -e {0}` — `errexit` only, **no `pipefail`**. Without `pipefail`, a shell pipeline observes only the LAST command's exit code. `tee` always exits `0`, so any non-zero from `cargo test` is silently swallowed and the step appears green.

The downstream `Assert bwrap SEC-002 regression tests ran (#1104)` step then fails with the misleading message:

```
::error::Required bwrap regression tests did not run:
  - git_config_deny_pre_creates_for_non_git_dir_to_close_toctou
  - git_config_deny_adds_ro_bind_for_existing_git_dir
  - ...
```

Real-world cost: ~30 min of debugging per occurrence. We hit this on **PR #1326** (run `25471442843`) AND **PR #1330** (run `25497008448`) — the second time even though we knew about it. Time to fix it once and for all.

## The fix (one line)

Add `shell: bash` to the Test step. Per the [GH Actions docs][1], `shell: bash` invokes the script via `bash --noprofile --norc -eo pipefail {0}` — pipefail enabled. Now `cargo test` failures fail the step they actually originated in, with the actual error message.

[1]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunshell

## Self-pin against future regression

To prevent someone from accidentally stripping `shell: bash` later (workflow cleanup pass, ignorance of the load-bearing context, etc.), added a tiny awk scan as a separate step in the same `fmt` job:

```yaml
- name: Pin pipefail on Test step (#1328)
  run: |
    if ! awk '/- name: Test$/{flag=1; next}
              flag && /shell: bash$/{found=1; exit}
              flag && /^      - name:/{exit}
              END{exit !found}' .github/workflows/ci.yml; then
      echo "::error::Test step in ci.yml is missing 'shell: bash' (#1328)."
      ...
      exit 1
    fi
```

Verified both directions:

| Scenario | Result |
|---|---|
| Current state (positive) | exits `0` → step passes |
| `sed`-stripped `shell: bash` (negative) | exits `1` → step fails with actionable error pointing at #1328 |

The pin lives **inside the workflow file it's protecting** — zero new files, zero new languages, zero new dependencies. Cheap (~50ms) and self-documenting.

### Why not a python/Rust regression test?

Per @lijunzh's pushback during the design discussion: a python regression test in a Rust repo smells weird, and a Rust integration test that greps a YAML file smells equally weird in reverse. The property being guarded is "this YAML file has the right text in the right place" — it should be guarded by YAML/CI tooling, in CI, by CI. The awk one-liner is exactly that.

## Bonus: bash port of `check_tokio_test_flavor.py`

While in the neighborhood, ported the one stray Python script in the workspace. Original was already a textual scan with no real Python-native logic — bash + grep + perl handle it cleanly.

### Design notes for the port

1. **No PCRE.** The original Python regex used a negative lookahead `#\[tokio::test\](?!\()` to match the bare form but not the flavored form. **The lookahead was redundant**: the literal `]` immediately after `test` already distinguishes bare from flavored (`#[tokio::test]` vs `#[tokio::test(...)]` — the latter never has `#[tokio::test]` as a substring). Dropping the lookahead lets us use plain ERE, which works on **both GNU grep (Linux CI) AND BSD grep (macOS local dev)**. The alternative was `grep -P` which BSD grep doesn't support — would have broken every macOS dev's local check.

2. **`git ls-files` over manual recursion.** Faster, and honors `.gitignore` for free so `target/` etc. are skipped without an explicit `SKIP_DIRS` list. Falls back to `find` if not in a git repo (rare).

3. **`perl -i` over `sed -i` for `--fix`.** GNU and BSD `sed` disagree on `-i` semantics (BSD requires an explicit empty string for "no backup suffix", GNU doesn't). Perl ships everywhere we run CI and uses identical regex semantics on both platforms.

4. **Output format byte-for-byte identical to the Python.** Any tooling that scrapes the output (none today, hypothetical future) keeps working.

### Verification (all four scenarios)

| Scenario | Result |
|---|---|
| Clean state | exits `0`, prints `[#1109 F2 guard] OK …` |
| Injected 2 bare `#[tokio::test]` into `koda-core/src/child_agent.rs` (real danger-pattern file) | exits `1`, lists offenders with line numbers + fix instructions |
| `--fix` on injected offenders | promoted 2/2 to `multi_thread`, exit `0`, file restored cleanly |
| `shellcheck scripts/check_tokio_test_flavor.sh` | clean (no warnings, no errors) |

### Reference cleanup

- `.github/workflows/ci.yml` — caller now invokes the `.sh`
- `koda-core/src/providers/http.rs` — docstring pointer updated
- `scripts/check_per_file_coverage.py` — docstring pointer updated
- `CHANGELOG.md` — **left untouched**: the historical entry referring to `scripts/check_tokio_test_flavor.py` is correct as history (that **was** the filename when the lint was originally added). Rewriting historical changelog entries is a smell.

## What's NOT in this PR

- **`scripts/check_per_file_coverage.py` not migrated.** That one parses JSON coverage data and computes per-file deltas — actually benefits from a real language. Bash + `jq` could do it but the script would be longer and harder to read. Python pulls its weight there. Separate scope if we ever want pure-Rust-only.
- **#1327 (the flake itself)** — separate PR, intentionally. With #1328 in place, future occurrences of #1327 will at least be diagnosed correctly instead of blaming bwrap.

## Risk

- **Surface area**: 1 YAML edit, 1 new bash script, 1 deleted Python script, 2 docstring pointer updates. No production code touched.
- **Rollback**: `git revert` is clean — the YAML edit and the script swap are independent commits' worth of changes but happen to share a single commit here. If only one half needs reverting, it's a 2-line diff each.

Refs: #1328 · related: #1109 (the lint this script implements), #1327 (the flake whose diagnosis #1328 unblocks)


lijunzh:
🎯 **Test (macos-latest) failure is a perfect proof-of-concept for this PR.**

The failure is `safe_trust_sub_agent_can_write_files ... FAILED` in `e2e_sub_agent_trust_test.rs:105` — that's #1327, the pre-existing flake we just opened the issue for.

**Critically**: notice WHERE the failure is reported. Pre-#1328, this exact same flake on PR #1326 (run `25471442843`) and PR #1330 (run `25497008448`) showed up as:

- ✓ Test (green!)
- ✗ "Required bwrap regression tests did not run" (the wrong step taking the blame)

…sending us down a 30-min security rabbit hole each time. With this PR's `shell: bash` (pipefail) fix:

- ✗ **Test (macos-latest)** — `safe_trust_sub_agent_can_write_files ... FAILED` ← actual cargo error, on the actual failing step

That's the entire point. 🧠 The fix works.

The flake itself will be addressed in the #1327 follow-up PR. Re-running the failing job to clear the queue for merge.

Issue #1250, opened 2026-05-04
Trust taxonomy: 3 modes (Plan/Safe/Auto) with context-sensitive 'ask' resolution; supersedes #1248 and #1249


lijunzh:
## Overview

Collapse the access-control machinery to **one mechanism**: `TrustMode` with 3 levels (Plan / Safe / Auto). The same enum applies in both top-level and sub-agent contexts, with a single context-sensitive rule that resolves the dead-channel problem #1249 surfaced.

This **supersedes #1249** (which was over-engineered with two separate enums) and **closes #1248** more cleanly than the SafeExec proposal.

## The 3-mode taxonomy

| | Read tools | Safe Bash (`ls`, `cat`, `git log`) | Mutating Bash (`cargo build`, `npm install`) | Write/Edit/Delete | Network (`curl`, `WebFetch`) | Destructive (`rm -rf`, `git reset --hard`, `git push --force`) |
|---|---|---|---|---|---|---|
| **Plan** | \u2705 auto | \u2705 auto | \u274c block | \u274c block | \u274c block | \u274c block |
| **Safe** | \u2705 auto | \u2705 auto | \u26a0\ufe0f ask | \u26a0\ufe0f ask | \u26a0\ufe0f ask | \u26a0\ufe0f ask |
| **Auto** | \u2705 auto | \u2705 auto | \u2705 auto | \u2705 auto | \u2705 auto | \u26a0\ufe0f ask |

Three rows, three plain English meanings:

1. **Plan** \u2014 read-only, no exceptions
2. **Safe** \u2014 read-only with safe writes, ask permission for anything mutating
3. **Auto** \u2014 approve almost everything, ask permission only for genuinely dangerous things

## The context-sensitive rule (resolves #1249's dead-channel discovery)

Sub-agents have no human at the keyboard (`sub_agent_dispatch.rs:173` creates a dead `cmd_rx` by design \u2014 #1022 B10). So *"ask"* means something different in a sub-agent than at top-level. **One rule resolves it:**

> **"Ask" means "default to safe-side."**
> - Top-level (human present): "ask" \u2192 prompt the user.
> - Sub-agent (no human): "ask" \u2192 do whatever is safer.

Applied per cell:

| Cell                          | Top-level         | Sub-agent     |
|-------------------------------|-------------------|---------------|
[...]

Issue #1249, opened 2026-05-04
Redesign trust taxonomy: split SessionTrust (top-level) from AgentCapability (sub-agents); supersedes #1248


lijunzh:
## Discovery

`koda-core/src/sub_agent_dispatch.rs:173`:

```rust
let (_, mut cmd_rx) = mpsc::channel(1);  // sender immediately dropped \u2192 dead channel
```

**Every sub-agent (including forks) gets a dead-end approval command channel.** When a sub-agent at `Safe` trust calls a `NeedsConfirmation` tool (Write, Edit, Delete, mutating Bash, curl, ...), `request_approval` waits on the dead channel, returns `None`, and the call site auto-rejects with *"requires user confirmation but this sub-agent has no channel to the user."* (See sub_agent_dispatch.rs:1480-1500.)

This is by design (#1022 B10): sub-agents are non-interactive; we don't want them pausing waiting for human input. **But it means our `TrustMode` taxonomy is misshapen for the sub-agent context.**

## What this means for sub-agents today

For sub-agent execution, the meaningful trust levels collapse to two:

| Declared trust | What sub-agent can actually do | Effective ceiling |
|---|---|---|
| `Plan`         | Read tools auto. ReadOnly Bash auto. Everything else `Blocked`. | **Read-only** |
| `Safe` (today) | Read tools auto. ReadOnly Bash auto. Everything else `auto-rejected: requires user confirmation`. | **Read-only** (same as Plan; only the error string differs) |
| `Auto`         | Everything auto-approved. | **Unrestricted** |

So today, `task.json` with `write_access: true` and implicit `Safe` trust **cannot actually write files** when invoked as a sub-agent. The model sees Write/Edit/Delete in its tool catalog, calls them, gets auto-rejected on every one, and must give up. The `write_access: true` is largely cosmetic for sub-agents.

The only way for a sub-agent to actually mutate things today is `Auto` \u2014 and Auto is "everything goes" with no granularity for *"writes yes, network no"* or vice versa.

## Why this matters more than #1248 (which this supersedes)

Issue #1248 proposed adding `SafeExec` for the "verify" use case (read + exec, no writes). But that fix only fills one gap in a fundamentally misshapen taxonomy. The real question discovered during #1248 review:

> *"Should worker (Safe) == read+exec (SafeExec)?"*

[...]

PR #1332, opened 2026-05-07
fix(test): deflake safe_trust_sub_agent_can_write_files (#1327)

### LINKED ISSUE #1327: Flaky: safe_trust_sub_agent_can_write_files races under workspace-parallel CI load

lijunzh:
## Summary

Closes #1327. Surgical deflake of `safe_trust_sub_agent_can_write_files` (and a consistency fix to its destructive sibling). One file touched, 31 lines added, 2 removed — most of the diff is rationale comments to prevent re-introduction.

## Root cause

The bg sub-agent's `MockProvider::from_env()` reads `KODA_MOCK_RESPONSES` **lazily**, on the bg task, **AFTER** the parent's `run_inference` returns. The previous test code did:

```rust
let events = env.run_inference(&provider).await;
runtime_env::remove("KODA_MOCK_RESPONSES");   // ← RACE
// ...10 lines of comment...
collect_bg_events_after(events, ...).await;   // ← wait happens here
```

That `remove()` ran **before** `collect_bg_events_after` confirmed the bg agent had constructed its provider. Under workspace-parallel load (CPU contention slows the bg task's first scheduling), the parent reliably wins the race:

1. Parent calls `runtime_env::set("KODA_MOCK_RESPONSES", <write script>)`
2. Parent's `run_inference` dispatches `InvokeAgent` → spawns bg agent task → returns immediately (per #1163, every `InvokeAgent` is now spawn-and-return)
3. **Parent calls `runtime_env::remove(...)`** ← bg agent hasn't been scheduled yet
4. Bg agent finally gets scheduled, calls `create_provider` → `MockProvider::from_env()` → reads **`[]`** (empty) instead of the configured Write script
5. Bg agent's first turn returns no tool call → terminal state (success!)
6. Parent's `collect_bg_events_after` returns happy (terminal state reached)
7. `assert!(target_path.exists())` panics with "sub-agent's Write must produce the file at ..." — and we go on a 30-min wild goose chase looking for a security regression that doesn't exist

## Fix

Move `runtime_env::remove(...)` to **after** `collect_bg_events_after`. Order matters:

| Step | Action |
|---|---|
| 1 | `set` `KODA_MOCK_RESPONSES` |
| 2 | spawn bg agent (via `run_inference` → `InvokeAgent` dispatch) |
| 3 | **WAIT** for bg agent to reach terminal state |
| 4 | **ASSERT** bg agent's side effects |
| 5 | **REMOVE** `KODA_MOCK_RESPONSES` |

By step 3, the bg agent has either consumed the env var or terminated without ever needing to. Either way, the `remove()` in step 5 can no longer affect bg agent behavior.

### 1. `safe_trust_sub_agent_can_write_files` (the actual flake)

Reordered. **Long-form rationale comment added in-line** to prevent re-introduction:

```
**#1327 deflake** — DO NOT remove KODA_MOCK_RESPONSES here.
The bg sub-agent's `MockProvider::from_env()` reads the env
var lazily on the bg task, AFTER the parent's `run_inference`
returns. ... Order matters: WAIT → ASSERT → REMOVE.
```

The next person to copy this test pattern will see WHY the order matters before they get bitten.

### 2. `safe_trust_sub_agent_destructive_bash_is_still_blocked` (consistency fix)

Same anti-pattern, but **"accidentally robust"** today: empty responses → no `rm -rf` → file survives → `target.exists()` assert passes vacuously. Fixed for consistency, because future readers will copy whichever pattern they see first — and the destructive test currently has the WRONG one. Also added the `collect_bg_events_after` wait it was missing entirely (was relying purely on `run_inference`'s parent-loop completion, which doesn't synchronize with the bg task at all).

## Why not migrate every `KODA_MOCK_RESPONSES` test?

Audited all 18 callsites in `koda-core/tests/`. They fall into safe categories:

| Pattern | Count | Why it's safe |
|---|---|---|
| **Foreground sub-agents** (no bg spawn) | many | No race possible. |
| **Use `invoke_agent_and_take_calls` helper** | 2 | Helper has `collect_bg_events_after` baked in (#1163). |
| **Use explicit `WaitTask`** in parent's mock script | 1 | Parent's `run_inference` blocks on bg-agent terminal state before returning. |
| **Test asserts only parent-side state** (lines 461, 1044) | 2 | Race exists in source-order sense but doesn't manifest as flake — nothing the test asserts depends on bg agent consuming the env var. |

Verdict: surgical fix. The acceptance-criteria carve-out for migrate-everything only applies if we change the env-var mechanism, which we don't.

### Local stress (M3 Pro, 14 cores)

| Run mode | Result |
|---|---|
| 80/100 isolated runs of the trust test | **0 fails** ✅ (matches the issue's 10/10 isolated baseline) |
| 2/2 full `cargo test -p koda-core --features test-support` runs | **0 fails** ✅ |
| Build clean (`cargo build -p koda-core --tests --features test-support`) | ✅ |
| `cargo fmt --check` | ✅ |

### CI canonical stress

The acceptance criteria asks for ≥100/100 under workspace-parallel CI on ubuntu-latest. The macOS Test job is the one that originally surfaced the flake (PR #1326 run `25471442843`, PR #1330 run `25497008448`). Both jobs will run on this PR — will retrigger a few times to build confidence.

## Risk

- **Surface area**: 1 test file, 31 added lines (mostly comments), 2 removed lines. **Zero production code touched.**
- **Rollback**: trivial `git revert`. The change is purely a statement reordering + new comments.
- **Could it deflake but introduce a new flake?** No — the new wait was already present in the file (just in the wrong order). All we did was move the `remove()` to after the wait that was already happening.

## Related

- **Closes**: #1327
- **Refs**: #1321 #1323 (prior partial deflake attempts that reduced rate from 4/5 to ~1/5 but didn't address root cause), #1250 #1249 (the bug this test pins), #1328 (the CI bug that masked this one and made it 10× harder to diagnose), #1163 (introduced the spawn-and-return InvokeAgent dispatch that created the race window in the first place)

### DIFF
--- a/koda-core/tests/e2e_sub_agent_trust_test.rs
+++ b/koda-core/tests/e2e_sub_agent_trust_test.rs
@@ -85,7 +85,18 @@ async fn safe_trust_sub_agent_can_write_files() {
         MockResponse::Text("Sub-agent reported: Done writing.".into()),
     ]);
     let events = env.run_inference(&provider).await;
-    runtime_env::remove("KODA_MOCK_RESPONSES");
+    // **#1327 deflake** — DO NOT remove KODA_MOCK_RESPONSES here.
+    // The bg sub-agent's `MockProvider::from_env()` reads the env
+    // var lazily on the bg task, AFTER the parent's `run_inference`
+    // returns. If we `remove` before `collect_bg_events_after`
+    // confirms the bg agent has consumed the responses, we race the
+    // bg agent and it sees `[]` instead of the configured script.
+    // Empty responses → no Write call → no file on disk → the
+    // `target_path.exists()` assert below panics. This is exactly
+    // the symptom #1327 documents. The previous "#1321/#1323
+    // deflake" wait below catches the bg agent's terminal state
+    // but couldn't help if the bg agent never even got the right
+    // script to act on. Order matters: WAIT → ASSERT → REMOVE.
 
     // **#1321/#1323 deflake**: this test was racing parent-loop
     // termination against bg-agent completion (the parent's Mock
@@ -98,6 +109,11 @@ async fn safe_trust_sub_agent_can_write_files() {
         .collect_bg_events_after(events.clone(), std::time::Duration::from_secs(10))
         .await
         .expect("bg sub-agent never reached terminal state");
+
+    // Now the bg agent has reached terminal state — if it consumed
+    // KODA_MOCK_RESPONSES at all, it did so already. Safe to clean
+    // up for the next test in the binary (#1327).
+    runtime_env::remove("KODA_MOCK_RESPONSES");
 
     // Assertion 1: the file actually exists on disk. This is the
     // strongest possible proof that the Write was approved and
@@ -186,7 +202,20 @@ async fn safe_trust_sub_agent_destructive_bash_is_still_blocked() {
         ),
         MockResponse::Text("Done.".into()),
     ]);
-    let _events = env.run_inference(&provider).await;
+    let events = env.run_inference(&provider).await;
+    // **#1327 deflake** — don't `runtime_env::remove` until AFTER
+    // we've waited for the bg sub-agent to reach terminal state.
+    // See the long-form rationale in the sibling test above. This
+    // test is "accidentally robust" to the race today (empty
+    // responses → no `rm -rf` → file survives → the file-survives
+    // assertion still passes vacuously), but the original
+    // `safe_trust_sub_agent_can_write_files` was "accidentally
+    // fragile". Same anti-pattern — fix both for consistency,
+    // because future readers will copy whichever they see first.
+    let _bg_events = env
+        .collect_bg_events_after(events.clone(), std::time::Duration::from_secs(10))
+        .await
+        .expect("bg sub-agent never reached terminal state");
     runtime_env::remove("KODA_MOCK_RESPONSES");
 
     // The destructive op MUST have been blocked. Filesystem state


lijunzh:
Test (ubuntu-latest) failure is **unrelated to this PR** \u2014 it's a cascade of `memory cache mutex poisoned: PoisonError` failures in `koda-core/src/memory.rs` unit tests. Root-cause panic: `test_cache_hit_skips_disk_reread` at `memory.rs:607:41` with \"path must be cached\". Looks like a parallel-test cache-contention flake (one test poisons the cache mutex, all subsequent tests cascade-fail with PoisonError).

This PR doesn't touch `memory.rs` or any of the failing tests \u2014 they're in a completely different module from the deflake fix.

For comparison: **Test (macos-latest) passed** (2m10s) on the same run, which is the platform that originally surfaced #1327. That's the canonical confidence signal for this PR.

Reran the failed job. Will file a separate issue for the memory-cache flake if it shows up again \u2014 it's a different bug class entirely (mutex poisoning under test parallelism, vs the env-var lazy-read race this PR fixes).

Issue #1349, opened 2026-05-08
Idle-state visibility & continuation gap: bg sub-agents go invisible and never re-engage the parent


lijunzh:
# Idle-state visibility & continuation gap: bg sub-agents go invisible and never re-engage the parent

## TL;DR

When a parent agent ends its turn cleanly (the model behaviour PR #1348 was designed to encourage), the framework treats the session as done. But background sub-agents continue running — and the framework has **no idle-state event/data path** to surface their activity or wake the parent when they finish. The result: the bottom pill freezes, mail piles up, the parent never re-engages, and the user assumes the session is broken and kills it.

This umbrella issue captures four interrelated bugs uncovered while debugging #1344 and validating PR #1348.

## Repro

Run any "spawn an explore sub-agent" prompt on **gemini-pro** (Pro is required — flash sidesteps this code path via fabrication, see Bug 4):

```
> quick explore the codebase by sub agents and help me understand what this repo is doing
```

Pro spawns the agent, writes a placeholder ("I'll let you know when the explore agent finishes"), and ends its turn cleanly (correct behaviour). Then:

1. Bottom-pill activity widget freezes on the first tool name (e.g. "List") even though the sub-agent does 30+ more tool calls. (Bug 1)
2. Sub-agent eventually completes and sends mail to the parent's mailbox. (Verified in conversation.md transcript via PR #1348's live persistence — proves data path is intact.)
3. Pill disappears (registry deregisters the completed task). (correct)
4. **Nothing else happens.** Parent remains idle forever. Mail never gets injected. Model never re-engages. (Bug 2)

Debug bundle: `koda-debug-20260508-130040-session.zip` (analyzed in #1348 PR thread).

Timeline from `logs/koda-56653.log`:

| Time | Event |
|---|---|
| 12:51:49 | Session starts |
| 12:54:30–12:55:12 | Sub-agent does 30+ tool calls |
| 12:55:12–12:58:49 | Sub-agent waits ~3.5 min for Pro to generate the final summary |
| **12:58:49** | `execute_sub_agent` span CLOSES — sub-agent COMPLETED |
| 12:58:49–13:00:40 | **Nothing**. Mail sits in mailbox. Parent idle. |
| 13:00:40 | User captures debug bundle, kills session |

The transcript captured all 35+ tool calls (PR #1348 live persistence works ✓). The TUI pill saw none of them after the parent's turn ended. The completion mail was never delivered to the model.

[...]

PR #1350, opened 2026-05-08
feat(peers): bootstrap script + manifest for cross-reference repos

lijunzh:
## What

Adds a checked-in bootstrap convention so reference clones of **Codex**, **Claude Code**, **Gemini CLI**, and **Zed** can be set up on any machine with one command:

```sh
./peers/bootstrap.sh
```

The script is idempotent and doubles as the daily sync command:

- Missing clone → `git clone` into `../<name>`
- Missing symlink → create `peers/<name>` → `../../<name>`
- Existing clone, clean tree → fast-forward to upstream HEAD
- Existing clone, dirty tree → skip (never touch your WIP)
- Diverged from upstream → skip (your problem to resolve)

## Why

We constantly pattern-match koda's design against how the same problem is solved in those four projects (mailbox plumbing, sub-agent dispatch, TUI event loops, etc.). Having their source one `rg peers/codex/codex-rs/ ...` away — without leaving the repo, without GitHub round-trips, with full IDE jump-to-definition — turns hours of context-switching into seconds.

Until today this was a single-user setup with un-tracked symlinks pointing at `../codex` etc. New machines (and future-me on a fresh laptop) had no way to reproduce it. This PR makes the convention reproducible without imposing anything on contributors who don't want it.

## Why **not** git submodules

I considered this and rejected it. Submodules **pin to specific commits**, but we want **always-at-HEAD** reference material:

| What we want | What submodules give us |
|---|---|
| Always at latest HEAD | Pinned to a specific commit; bumping = noisy "chore: bump peers" commits |
| Zero cost for koda installation | `git clone --recurse-submodules` pulls **gigabytes** (zed alone is huge) |
| Personal reference, not part of koda | `.gitmodules` publicly declares codex/zed as koda dependencies — they aren't |
| Update independently of koda commits | The whole point of submodules is that they DON'T |

Plus the usual submodule footguns (detached HEAD on update, "submodule is dirty" warnings, `--recurse-submodules` flag everyone forgets).

## Layout

```
peers/
├── README.md          ← checked in (explains the design)
├── peers.txt          ← checked in (manifest: name + git URL)
├── bootstrap.sh       ← checked in (clone-or-update + symlink)
├── .gitkeep           ← checked in (keeps the dir present)
└── <name>/            ← gitignored (symlink to ../../<name>)
```

The clones live **outside** the koda working tree so they never count against koda's clone size, never trigger `cargo`/`npm`/IDE indexers that walk the workspace, and can be `git pull`-ed independently.

The `.gitignore` flips from `peers/` to `peers/*` with an allowlist for the four meta-files. Adding a new peer is one line in `peers.txt` + a re-run of `bootstrap.sh` — no script edits needed.

## Tested locally

- Bootstrap on existing clones → no-op + fast-forwards as expected (`zed` picked up 2 fresh upstream commits during the test run)
- Bootstrap with deliberately-missing symlink → recreates it
- `shellcheck peers/bootstrap.sh` → clean (exit 0)
- `git status` after staging → only the four meta-files appear; symlinks invisible
- Pre-push clippy hook → passed

## Backstory / context for reviewers

This came out of the #1349 debugging session — pulling and updating the four reference repos by hand to compare implementations was painful enough that "let's formalize this" became obvious. The original quick-and-dirty version (symlinks + a local `sync.sh`) is what's being replaced here; both predecessors lived under a `peers/` ignore from minute one, so nothing's leaking from prior states.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,14 @@ docs/book/
 # Python bytecode (scripts/ contains a few helper scripts)
 __pycache__/
 *.pyc
+
+# peers/ — local symlinks + clones to reference projects (Codex, Claude
+# Code, Gemini CLI, Zed) used during koda development. The meta-files
+# (manifest, bootstrap script, README) are checked in so contributors
+# can run `./peers/bootstrap.sh` on a fresh laptop; the symlinks and
+# clones stay strictly local. See peers/README.md for the design.
+peers/*
+!peers/.gitkeep
+!peers/README.md
+!peers/peers.txt
+!peers/bootstrap.sh
--- a/peers/README.md
+++ b/peers/README.md
@@ -1,0 +1,88 @@
+# `peers/` — cross-reference repos for koda development
+
+This directory holds **symlinks to sibling clones of peer projects** we
+study while building koda — Claude Code, Codex, Gemini CLI, and Zed.
+
+## Why this exists, in one paragraph
+
+When we design a feature (mailbox plumbing, sub-agent dispatch, TUI
+event loop, etc.) we constantly pattern-match against how the same
+problem is solved in those four projects. Having their source one
+`rg peers/codex/codex-rs/ ...` away — without leaving the repo, without
+GitHub round-trips, with full IDE jump-to-definition — turns hours of
+context-switching into seconds.
+
+## Why **not** git submodules
+
+Submodules **pin to specific commits**. We want **always-at-HEAD**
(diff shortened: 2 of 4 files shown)


lijunzh:
Auto-closed by branch deletion. Replaced by #1351 — same content, with the script renamed from `bootstrap.sh` to `sync.sh` (see naming-note section in the new PR for the rationale).

